PR #14, opened 2024-06-24
single anvil instance

hamdiallam:
worked on this with @jakim929.

Spins up a single vanilla anvil instance. Gives us a base to creating more than one instance

### DIFF
--- a/supersim.go
+++ b/supersim.go
@@ -19,9 +19,9 @@ func NewSupersim(log log.Logger) *Supersim {
 	return &Supersim{log, anvil}
 }
 
-func (s *Supersim) Start(_ context.Context) error {
+func (s *Supersim) Start(ctx context.Context) error {
 	s.log.Info("starting supersim")
-	return s.anvil.Start()
+	return s.anvil.Start(ctx)
 }
 
 func (s *Supersim) Stop(_ context.Context) error {
@@ -30,5 +30,5 @@ func (s *Supersim) Stop(_ context.Context) error {
 }
 
 func (s *Supersim) Stopped() bool {
-	return false
+	return s.anvil.Stopped()
 }
--- a/supersim.go
+++ b/supersim.go
@@ -3,25 +3,30 @@ package supersim
 import (
 	"context"
 
+	"github.com/ethereum-optimism/supersim/anvil"
+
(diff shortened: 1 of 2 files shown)


supersim.go
```diff
@@ -3,25 +3,30 @@ package supersim
...
 }
 
-func (s *Supersim) Stop(ctx context.Context) error {
+func (s *Supersim) Stop(_ context.Context) error {
 	s.log.Info("stopping supersim")
-	return nil
+	return s.anvil.Stop()
 }
 
 func (s *Supersim) Stopped() bool {
```
fainashalts, line 32: This should probably return `s.anvil.Stopped()` maybe instead of false, to reflect the status of the anvil instance. What do you think?

hamdiallam: ah yea that's right! switched the lifecycle a bit since the last draft and didn't update this. nice catch

fainashalts:
So exciting to see y'all worked on this today! 🎉 Thank you for taking the initiative. I think this will form a great base for us to work off of tomorrow. I had one question about context. Do we want to propagate the context from supersim to anvil in the orchestrator? If I understand correctly, doing so would ensure that any cancellations or timeouts applied to the supersim context would also affect the anvil context. I'm thinking specifically of avoiding the scenario where we get an orphan process if supersim gets shut down but anvil keeps running. What do you think?

https://medium.com/@jamal.kaksouri/the-complete-guide-to-context-in-golang-efficient-concurrency-management-43d722f6eaea

hamdiallam:
@fainashalts 

So originally we did propagate the context but I thought it would be better to follow the `cliapp.Lifecycle` interface that the `Supersim` struct implements to be explicit. The cliapp handles signaling interrupt information of the parent process through the `Start` & `Stop` methods. So from that, we only need to make sure everything winds down on `Stop`

hamdiallam:
actually can do both! @fainashalts 

added a listener to also cancel the resource context if the threaded through context on `Start` is done

fainashalts:
> actually can do both! @fainashalts
> 
> added a listener to also cancel the resource context if the threaded through context on `Start` is done

Nice! Best of both worlds. 😄

PR #27, opened 2024-06-26
Adds basic tic tac toe game

nitaliano:
**Description**

* Adds basic `TicTacToe` contract along with unit tests. This is a test contract we're going to continue to evolve to demonstrate how interop works.
* Removes the counter contracts

**How to play**
* Player1 calls `createGame` this returns a game id
* Player2 calls `joinGame` with the game id
* Player1 has the first turn, then they rotate until a draw or game is won, the players call `makeMove` with game id and spot on the board.


contracts/src/TicTacToe.sol
tremarkley, line 1: this might be overkill, but should we have verification that makes sure the address making a move or creating the game is the actual owner of that address? with the way the game is set up right now, anyone can make a move or create a game on behalf of any address.

tremarkley (commented):
Nice!

contracts/src/TicTacToe.sol
nitaliano: Yea i did it this way intentionally, because I'm going to try and get the FE to also work with smart accounts and just wanted it to be easier to setup at first.

tremarkley (approved):

PR #31, opened 2024-06-28
feat: add startup logging

jakim929:
waits until the anvil nodes are ready to accept requests, then prints chain id and RPC configs

```
Supersim Config:
L1:
  Chain ID: 1    RPC: http://127.0.0.1:8545
L2:
  Chain ID: 10    RPC: http://127.0.0.1:9545
  Chain ID: 30    RPC: http://127.0.0.1:9555
```


jakim929:
* **#31** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/31?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

anvil/anvil.go
```diff
@@ -109,11 +106,9 @@ func (a *Anvil) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to start anvil: %w", err)
 	}
 
-	if _, err := utils.WaitForAnvilClientToBeReady(fmt.Sprintf("http://%s:%d", host, a.cfg.Port), 5*time.Second); err != nil {
-		return fmt.Errorf("failed to start anvil: %w", err)
-	}
-
 	go func() {
+		defer os.Remove(tempFile.Name())
```
tremarkley, line 110: oh interesting! so moving `defer os.Remove(tempFile.Name())`, makes it so we no longer need to use `utils.WaitForAnvilClientToBeReady`? I'm curious since the call to `defer os.Remove(tempFile.Name())` is now here, is it possible that if the process errored out prior to this, would the `tempFile` not end up being removed?

utils/rpc.go
```diff
@@ -3,13 +3,27 @@ package utils
...
-	"net/http"
+	"strings"
 	"time"
 
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
-func WaitForAnvilClientToBeReady(rpcUrl string, timeout time.Duration) (*rpc.Client, error) {
+func WaitForAnvilEndpointToBeReady(endpoint string, timeout time.Duration) error {
+	client, clientCreateErr := rpc.Dial(endpoint)
```
hamdiallam, line 13:
```suggestion
	client, err := rpc.Dial(endpoint)
```

tremarkley (approved):

utils/rpc.go
hamdiallam: errs are very rarely specifically named

utils/rpc.go
```diff
@@ -3,13 +3,27 @@ package utils
...
+	if clientCreateErr != nil {
+		return fmt.Errorf("failed to create client: %v", clientCreateErr)
+	}
+
+	err := WaitForAnvilClientToBeReady(client, timeout)
```
hamdiallam, line 18:
```suggestion
	if err := WaitForAnvilClientToBeReady(client, timeout); err != nil {
```

nit

supersim.go
```diff
@@ -55,12 +59,27 @@ func (s *Supersim) Start(ctx context.Context) error {
 		return fmt.Errorf("l1 chain failed to start: %w", err)
 	}
 
+	var wg sync.WaitGroup
+	waitForAnvil := func(anvil *anvil.Anvil) {
+		defer wg.Done()
+		wg.Add(1)
+		utils.WaitForAnvilEndpointToBeReady(anvil.Endpoint(), 10*time.Second)
```
hamdiallam, line 66: we should avoid generic packages like `util` unless necessary.

How about adding a `Ready()` function to the `Anvil` struct. Since the implementation is domain specific anyways

hamdiallam (commented):
nice work! Left a few comments

anvil/anvil.go
jakim929: the behavior is slightly different than before 
- prev: as soon as anvil is ready we delete the temp file
- now: temp file is deleted when the process ends / errors out

mostly made this change so that the start function doesn't block - it returns immediately

utils/rpc.go
jakim929: got it! will fix

supersim.go
jakim929: got it! just refactored the code so that both supersim & anvil have a `WaitUntilReady` function and got rid of the util package entirely

Issue #16, opened 2024-06-25
op-simulator: basic proxy


fainashalts:
The prerequisite to additional work on the op-simulator is to add a basic proxy that takes RPC requests from a developer's app and pass them along to the local nodes. This initial task is to create this proxy as essentially only a pass through for now. 

**Definition of Done**

- [ ] `op-simulator` package exists
- [ ] `op-simulator` can be started from the `orchestrator`
- [ ] `op-simulator` intercepts RPC calls from a user app and passes all of these calls to identified local nodes

PR #33, opened 2024-07-08
feat: create op-simulator as basic proxy

tremarkley:
Closes https://github.com/ethereum-optimism/supersim/issues/16

This PR makes the following changes:
- creates `op_simulator` package
- sets up `op_simulator` for each chain as a ReverseProxy that intercepts all rpc requests and forwards them to the underlying anvil node for that chain
- coordinates starting and stopping of all `op_simulator` and `anvil` instances using the `supersim` package

### DIFF
--- a/op-simulator/op_simulator.go
+++ b/op-simulator/op_simulator.go
@@ -1,0 +1,99 @@
+package op_simulator
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"strconv"
+	"sync/atomic"
+
+	ophttp "github.com/ethereum-optimism/optimism/op-service/httputil"
+	"github.com/ethereum-optimism/supersim/anvil"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type Config struct {
+	Port uint64
+}
+
+type OpSimulator struct {
+	log        log.Logger
+	anvil      *anvil.Anvil
+	httpServer *ophttp.HTTPServer
+
+	stopped atomic.Bool
+
+	cfg *Config
+}
+
+const (
+	host = "127.0.0.1"
+)
+
+func New(log log.Logger, cfg *Config, anvil *anvil.Anvil) *OpSimulator {
+	return &OpSimulator{
+		log:   log,
+		cfg:   cfg,
+		anvil: anvil,
+	}
+}
+
+func (opSim *OpSimulator) Start(ctx context.Context) error {
+	proxy, err := opSim.createReverseProxy()
+	if err != nil {
+		return fmt.Errorf("error creating reverse proxy: %w", err)
+	}
+	mux := http.NewServeMux()
+	mux.Handle("/", proxy)
+	endpoint := net.JoinHostPort(host, strconv.Itoa(int(opSim.cfg.Port)))
+
+	hs, err := ophttp.StartHTTPServer(endpoint, mux)
+	if err != nil {
+		return fmt.Errorf("failed to start HTTP RPC server: %w", err)
+	}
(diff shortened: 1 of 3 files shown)


tremarkley:
* **#33** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/33?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

op-simulator/op_simulator.go
```diff
@@ -0,0 +1,99 @@
...
+	mux := http.NewServeMux()
+	mux.Handle("/", proxy)
+	endpoint := net.JoinHostPort(host, strconv.Itoa(int(opSim.cfg.Port)))
+
+	hs, err := ophttp.StartHTTPServer(endpoint, mux)
+	if err != nil {
+		return fmt.Errorf("failed to start HTTP RPC server: %w", err)
+	} else {
+		opSim.log.Info(fmt.Sprintf("listening on %v", endpoint), "chain.id", opSim.ChainId())
+	}
```
nitaliano, line 59: you can kill the else here and just do the log after the if statement

```
if err != nil {
  return fmt.Errorf("failed to start HTTP RPC server: %w", err)
}
opSim.Log.Info(fmt.Sprintf("listening on %v", endpoint), "chain.id", opSim.ChainId())
```

supersim.go
```diff
@@ -25,41 +31,54 @@ var genesisL1JSON []byte
...
+	l1Chain:  Chain{anvilConfig: anvil.Config{ChainId: 1, Port: 8545, Genesis: genesisL1JSON}, opSimConfig: op_simulator.Config{Port: 8546}},
+	l2Chains: []Chain{{anvilConfig: anvil.Config{ChainId: 10, Port: 9545, Genesis: genesisL2JSON}, opSimConfig: op_simulator.Config{Port: 9546}}, {anvilConfig: anvil.Config{ChainId: 30, Port: 9555, Genesis: genesisL2JSON}, opSimConfig: op_simulator.Config{Port: 9556}}},
 }
 
 type Supersim struct {
 	log log.Logger
 
-	l1Chain  *anvil.Anvil
-	l2Chains map[uint64]*anvil.Anvil
+	l1Anvil  *anvil.Anvil
```
hamdiallam, line 41: Nice work!

In order to simplify the changes in this file, I wonder if we should simply have OpSimulator manage the lifecycle of the the underlying chain, in this case anvil. So that we dont need reference to each individual pieces here

Not worth addressing in the PR. Just some food for thought for ways to clean up this code in the future

hamdiallam (approved):

op-simulator/op_simulator.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+type Config struct {
+	Port uint64
+}
+
+type OpSimulator struct {
+	log        log.Logger
+	anvil      *anvil.Anvil
+	HttpServer *ophttp.HTTPServer
```
hamdiallam, line 26: any reason `HttpServer` is exposed as a public field? Seems like it's not used elsewhere

supersim.go
tremarkley: Sounds good, I wont address in this PR, since I want to keep this change small. 

If we want to stick to the [original design](https://github.com/ethereum-optimism/design-docs/blob/7fce2c9e51bdd180c29d8805331905861d6c7c51/ecosystem/supersim.md#architecture), we should create an `Orchestrator` service that handles spinning up all the anvil instances and `OpSimulator` instances.

op-simulator/op_simulator.go
tremarkley: whoops, should not be public

PR #34, opened 2024-07-09
supersim testsuite creator

hamdiallam:
a test suite wrapper over a running instance of supersim. This makes it DRY between tests.

This wrapper will evolve as tests are added

Also introduces a dependency on stretchr/testify to write cleaner tests

### DIFF
--- a/supersim_test.go
+++ b/supersim_test.go
@@ -2,14 +2,13 @@ package supersim
 
 import (
 	"context"
-	"fmt"
-	"log"
-	"os"
 	"testing"
 	"time"
 
-	oplog "github.com/ethereum-optimism/optimism/op-service/log"
+	"github.com/ethereum-optimism/optimism/op-service/testlog"
+	"github.com/stretchr/testify/require"
 
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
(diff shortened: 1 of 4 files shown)


supersim_test.go
```diff
@@ -21,57 +20,70 @@ const (
...
-
-	defer func() {
-		err := supersim.Stop(context.Background())
-		if err != nil {
-			t.Fatalf("Failed to stop supersim: %v", err)
+type TestSuite struct {
+	t *testing.T
+
+	Cfg      *Config
+	Supersim *Supersim
```
tremarkley, line 27: is there a reason that these fields should be public?

supersim_test.go
```diff
@@ -21,57 +20,70 @@ const (
...
+}
 
-		if clientCreateErr != nil {
-			t.Fatalf("Failed to create client: %v", clientCreateErr)
-		}
+func TestStartup(t *testing.T) {
+	testSuite := createTestSuite(t)
+
+	// test that all chains can be queried
+	client, err := rpc.Dial(testSuite.Supersim.l1Chain.Endpoint())
```
tremarkley, line 57: just a heads up, once you merge in latest changes on master branch, make sure to update this so that the endpoint for these tests points to the op simulator endpoint and not directly to the anvil endpoint, so that we can test to make sure that the proxy is working properly (i.e. `testSuite.Supersim.l1OpSim.Endpoint()`)

tremarkley (approved):

supersim_test.go
```diff
@@ -21,57 +20,70 @@ const (
...
+	require.NoError(t, err)
+	require.NoError(t, client.CallContext(context.Background(), nil, "eth_chainId"))
```
tremarkley, line 59: Is it worth checking that the value of the chain id in the response is the same as the chain id in the l1 chain config?

supersim_test.go
```diff
@@ -21,57 +20,70 @@ const (
...
+	client.Close()
+
+	for _, l2Chain := range testSuite.Supersim.l2Chains {
+		client, err := rpc.Dial(l2Chain.Endpoint())
+		require.NoError(t, err)
+		require.NoError(t, client.CallContext(context.Background(), nil, "eth_chainId"))
```
tremarkley, line 65: similar comment here, Is it worth checking that the value of the chain id in the response is the same as the chain id in the corresponding l2 chain config?

tremarkley (approved):

supersim_test.go
hamdiallam: yep good shout

supersim_test.go
hamdiallam: For now they could technically be private. I made them public in the even this test suite is used in a diff package

supersim_test.go
hamdiallam: ah yea why not, i'll add

fainashalts (approved):
Nice! Thanks for getting our test suite started! 🎉

PR #36, opened 2024-07-09
feat: starting with port 0

hamdiallam:
We dont want to have to specify both the anvil & simulator ports when the anvil port is an implementation detail.

In order to simplify config, we provide the ability to start anvil on port `0`, meaning whatever is available on the host. This way we don't need to predefine this port. Since anvil is an external process, we extract the port by waiting for the `Listening on` log emitted when starting anvil.

## Fixes
Only apply the genesis arg to anvil if present in the config

### DIFF
--- a/anvil/anvil.go
+++ b/anvil/anvil.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -38,7 +39,8 @@ type Anvil struct {
 }
 
 const (
-	host = "127.0.0.1"
+	host                 = "127.0.0.1"
+	anvilListeningLogStr = "Listening on"
 )
 
 func New(log log.Logger, cfg *Config) *Anvil {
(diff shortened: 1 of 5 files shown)


anvil/anvil.go
```diff
@@ -58,22 +60,24 @@ func (a *Anvil) Start(ctx context.Context) error {
...
+	if len(a.cfg.Genesis) > 0 {
+		tempFile, err := os.CreateTemp("", "genesis-*.json")
+		if err != nil {
+			return fmt.Errorf("error creating temporary genesis file: %w", err)
+		}
+		if _, err = tempFile.Write(a.cfg.Genesis); err != nil {
+			return fmt.Errorf("error writing to genesis file: %w", err)
+		}
+
+		defer os.Remove(tempFile.Name())
```
hamdiallam, line 79: After anvil starts, we don't need this tempfile. Can remove immediately when this function returns. It's consumed in memory on anvil startup

supersim.go
```diff
@@ -27,8 +27,8 @@ var genesisL2JSON []byte
 var DefaultConfig = Config{
 	l1Chain: anvil.Config{ChainId: 1, Port: 8545, Genesis: genesisL1JSON},
```
tremarkley, line 28: nit: I think we should also set the port here to 0, since we cannot guarantee that 8545 is available when running tests.

tremarkley (approved):

PR #42, opened 2024-07-10
anvil logs to a tmp file

hamdiallam:
for now we rely on inspect anvil logs to grab the port when bound to `0`.

Rather than use the `--silent` flag, dump anvil log to temporary files such that they are inspectable for the user

### DIFF
--- a/anvil/anvil.go
+++ b/anvil/anvil.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strconv"
@@ -26,7 +27,8 @@ type Config struct {
 type Anvil struct {
 	rpcClient *rpc.Client
 
-	log log.Logger
+	log         log.Logger
+	logFilePath string
 
 	cfg *Config
 	cmd *exec.Cmd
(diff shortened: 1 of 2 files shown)


supersim.go
```diff
@@ -204,11 +196,11 @@ func (s *Supersim) ConfigAsString() string {
...
 	fmt.Fprintf(&b, "\nSupersim Config:\n")
 	fmt.Fprintf(&b, "L1:\n")
-	fmt.Fprintf(&b, "  Chain ID: %d    RPC: %s\n", s.l1OpSim.ChainId(), s.l1OpSim.Endpoint())
+	fmt.Fprintf(&b, "  Chain ID: %d     RPC: %s    LogPath: %s\n", s.l1OpSim.ChainId(), s.l1OpSim.Endpoint(), s.l1Anvil.LogPath())
 
 	fmt.Fprintf(&b, "L2:\n")
-	for _, l2OpSim := range s.l2OpSims {
-		fmt.Fprintf(&b, "  Chain ID: %d    RPC: %s\n", l2OpSim.ChainId(), l2OpSim.Endpoint())
+	for id, l2OpSim := range s.l2OpSims {
+		fmt.Fprintf(&b, "  Chain ID: %d    RPC: %s    LogPath: %s\n", l2OpSim.ChainId(), l2OpSim.Endpoint(), s.l2Anvils[id].LogPath())
```
nitaliano, line 203: i doubt this will happen the way everything is setup but i'm wondering if we should add a `nil` check for this specifically `s.l2Anvils[id].LogPath()`

anvil/anvil.go
```diff
@@ -90,7 +92,12 @@ func (a *Anvil) Start(ctx context.Context) error {
 	anvilPortCh := make(chan uint64)
 
 	// Handle stdout/stderr
-	//  - TODO: Figure out best way to dump into logger. Some hex isn't showing appropriately
+	logFile, err := os.CreateTemp("", fmt.Sprintf("anvil-chain-%d-", a.cfg.ChainId))
```
nitaliano, line 95: extra `-` at the end?

nitaliano (approved):

supersim.go
hamdiallam: hmm given the setup there's no good way to gracefully recover from a programming mistake here so better to simply panic

we can add a unit test though that asserts the key is present in both maps when setup. That should cover this case

anvil/anvil.go
hamdiallam: ah wait no actually the extra `-` is there becase `CreateTemp` appends a random string of characters at the end for name collision

Issue #6, opened 2024-06-24
orchestrator: configuration & vanilla mode


fainashalts:
There are two modes that supersim can run in. By default, vanilla is the mode of operation

# Vanilla

```
$ supersim // default vanilla mode
```

**There's no configuration**. We assert in the docs that supersim has some ports that it uses:
 * L1 Port: 8545
 * L2 Port(s): 9545,9546,954.. incrementing
 
Chain IDs are also predetermined:
* L1: 420
* L2(s): 991,992,993...

Vanilla mode run a single "l1" instance and two "l2" instances.

## Forked

**not implemented in this ticket**. We simply define the configuration story here in comparison to vanilla. A subcommand `fork` is available to fork from the superchain registry

TODO: create ticket for this command

```
$ supersim fork
```

By default latest L1 height. A specified height can be set as well
```
--fork.l1.height 420
```

By default, the mainnet eth network is forked. A testnet can be specified
```
--network sepolia
```

By default, the entire registry is forked. A subset of the superchain registry can be forked by name, delineated by commas. Fails if a name is not recognized in the registry
```
--chains "mainnet,base,zora.."
```

By default the fork urls will use the public endpoints but this should be allowed to be specified
```
--fork.urls "base=http://foo.bar,zora=http://bar.baz"
```

**Definition of Done**

Vanilla mode is operational. No configuration options. Simply run the supersim command.

Issue #25, opened 2024-06-25
Orchestrator: Fork mode


hamdiallam:
A follow up from #6 that implements the fork mode.

1. Add superchain registry as a dependency
2. Validate the fork configuration supplied from the user
3. From the L1 height, find the latest L2 heights for all specified networks
4. Skip any genesis application as remote state already has required contracts.
5. Forward fork information to each anvil instance

PR #43, opened 2024-07-10
feat(orchestrator): create orchestrator package

tremarkley:
Closes https://github.com/ethereum-optimism/supersim/issues/6

Creates the `orchestrator` package with the following properties:
- no config or arguments are read in and `orchestrator` spins up a default configuration of chains
  - fork mode support will be added in https://github.com/ethereum-optimism/supersim/issues/25
- refactors `supersim` package to use `orchestrator` and moves all startup code for `op-simulator` and `anvil` into the `orchestrator` package


tremarkley:
* **#43** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/43?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

orchestrator/orchestrator.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, fmt.Errorf("supersim does not support more than one l1")
+	}
+
+	for _, chainConfig := range config.ChainConfigs {
+		var genesis []byte
+		if chainConfig.BaseChainId == 0 {
+			genesis = genesisL1JSON
+		} else {
+			genesis = genesisL2JSON
+		}
```
nitaliano, line 62: this can be condensed slightly, something like this

```
genesis := genesisL2JSON
if chainConfig.BaseChainId == 0 {
  genesis = genesisL1JSON
}
```

op-simulator/op_simulator.go
```diff
@@ -22,7 +22,8 @@ const (
 )
 
 type Config struct {
-	Port uint64
+	Port        uint64
+	BaseChainId uint64
```
nitaliano, line 26: `SourceChainId` is also an option for this it's what we call it in viem if you wanted to follow a similar pattern. Another thing when it comes to naming variables/functions with `Id` in it you'll normally see it written as `BaseChainID`

orchestrator/orchestrator.go
```diff
@@ -0,0 +1,213 @@
...
+
+	handleErr := func(e error) {
+		if e != nil {
+			once.Do(func() {
+				err = e
+				cancel()
+			})
+		}
+	}
+
```
nitaliano, line 145: to reduce the nesting we can also write this block like

```
handleErr := func(e error) {
    if e == nil {
	return
    }

    once.Do(func() {
        err = e
	cancel()
    })
}
```

orchestrator/orchestrator.go
```diff
@@ -0,0 +1,213 @@
...
+	return true
+}
+
+func (o *Orchestrator) WaitUntilReady() error {
+	var once sync.Once
+	var err error
+	ctx, cancel := context.WithCancel(context.Background())
```
nitaliano, line 135: I'm wondering if we should consider adding a timeout at all https://pkg.go.dev/context#WithTimeout

orchestrator/orchestrator.go
```diff
@@ -0,0 +1,213 @@
...
+	return result
+}
+
+func (o *Orchestrator) ConfigAsString() string {
+	var b strings.Builder
+
+	fmt.Fprintf(&b, "\nSupersim Config:\n")
+	if o.L1OpSimulator() != nil {
+		fmt.Fprintf(&b, "L1:\n")
+		fmt.Fprintf(&b, "  Chain ID: %d    RPC: %s\n", o.L1OpSimulator().ChainId(), o.L1OpSimulator().Endpoint())
```
nitaliano, line 202: we could consider encapsulating these strings if we wanted so it can just be something like `o.L1OpSimulator().String()`

nitaliano (approved):
Good job! I'm starting to think you've been writing go longer than you're saying 😄

orchestrator/orchestrator.go
tremarkley: @nitaliano I think for now we dont need one since this is only waiting on anvil instances to be ready, and that wait function already has a timeout in it: https://github.com/ethereum-optimism/supersim/blob/e556e6e296dea56a6ef5b4b23ae4d250477a5c07/anvil/anvil.go#L204. In the future it could be necessary if there are other services we need to wait on. I'll leave it out for now since we don't currently need it. Let me know if you think otherwise!

Issue #21, opened 2024-06-25
Orchestrator: chain-config-loader


fainashalts:
Default chain configs exist in the Superchain registry, and in forking mode it makes sense to acquire these configs and use them as defaults for the L2 chains that `orchestrator` spins up.

To complete this task, we also need to decide how the default chain configs flow with any user-provided configs. If a user provides their own configs, do we ignore the registry configs? Are there variables that must be just so for particular superchain features to work? Part of this task is determining the answers to these questions and aligning the team on the planned solution. 

**Definition of Done**
The `chain-config-loader` acquires and applies relevant chain configs from the Superchain registry. A flow for how these configs interface with user-provided configs has been decided upon and applied.

Issue #20, opened 2024-06-25
orchestrator: fork state finder


fainashalts:
In order to work for forking, `orchestrator`'s l2-fork-state-finder` must acquire the L1 reference block at time of forking, and use that to determine the L2 block height and apply it to the relevant L2 instances.

**Definition of Done**
`orchestrator` is able to determine L2 block height and apply it to the L2 Anvil instances it spins up.

PR #51, opened 2024-07-16
supersim: fork mode

hamdiallam:
Closes ethereum-optimism/supersim#25
Closes ethereum-optimism/supersim#20
Closes ethereum-optimism/supersim#21

Pulls in ethereum-optimism/superchain-registry as a dependency and sets up the `fork` command. The orchestrator configuration can be generated from the supplied L1 fork height (defaults to the latest L1 state) and threads through fork configuration to anvil. The genesis application is omitted when fork mode is used 

There's some clean up to do with the different config structs but will leave that to followup PRs

## Summary by CodeRabbit

- **New Features**
  - Introduced "Forked mode" with instructions on forking chains in a superchain network, specifying parameters like L1 fork height and chains to fork.

- **Documentation**
  - Updated README with detailed instructions and help text for the `supersim fork` command.


orchestrator/fork.go
```diff
@@ -0,0 +1,105 @@
...
+	// L1
+	l1Client, err := ethclient.Dial(superchain.Config.L1.PublicRPC)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial l1 public rpc: %w", err)
+	}
+
+	l1ForkHeight := big.NewInt(rpc.FinalizedBlockNumber.Int64())
+	if cfg.L1ForkHeight > 0 {
+		l1ForkHeight = new(big.Int).SetUint64(cfg.L1ForkHeight)
+	}
```
tremarkley, line 31: should this come before `rpc.FinalizedBlockNumber`? that way if `cfg.L1ForkHeight` is set we avoid having to make the rpc for the finalized block number?

orchestrator/fork.go
```diff
@@ -0,0 +1,105 @@
...
+		Mnemonic:       DefaultMnemonic,
+		DerivationPath: DefaultDerivationPath.String(),
+		ForkConfig: &ForkConfig{
+			RPCUrl:      superchain.Config.L1.PublicRPC,
+			BlockNumber: l1Header.Number.Uint64(),
+		},
+	})
+
+	// L2s
+	l2ChainIds := cfg.Chains
```
tremarkley, line 50: is it possible the l1 chain id could end up in `cfg.Chains` or would that be caught by the `isChainInSuperchain` check?

supersim_test.go
tremarkley, line 1: are there any tests you can think of that might be useful for fork mode?

cmd/main.go
tremarkley, line 1: Now that we have a README, can you add documentation for this new functionality to it?

tremarkley (approved):

orchestrator/fork.go
hamdiallam: so we don't actually end up making an extra rpc here!

this just sets up the label to be used for the rpc on line 32. If `cfg.L1ForkHeight` is set, it'll override `l1ForkHeight` from the finalized label to the block number

orchestrator/fork.go
hamdiallam: yep exactly it would be caught in the CLI validation step since it'll be missing the superchain list of chains

cmd/main.go
hamdiallam: yep sounds good

PR #52, opened 2024-07-17
supersim: config refactors

hamdiallam:
Not every package needs to define it's external configuration and can instead be shared.

1. Expand on the `config` package and fold `chainapi` into config
2. Make startup start/stop logs less noisy. Switch a lot of detail logs to `debug` so that they are available if needed

## Summary by CodeRabbit

- **New Features**
  - Introduced WebSocket endpoint support.
  - Added name handling for configurations and simulator components.

- **Bug Fixes**
  - Improved logging with more detailed information.

- **Refactor**
  - Consolidated configuration handling into a single `ChainConfig` structure.
  - Updated method signatures to accept `ChainConfig` instead of individual parameters.
  - Simplified error handling and initialization logic in multiple components.
  
- **Tests**
  - Adjusted test configurations to align with the refactored `ChainConfig` structure.
  
- **Chores**
  - Removed unused imports and variables for cleaner codebase.


jakim929 (approved):
lgtm!

I do think the chainApi stuff should live in a separate package - since it's not exactly "config". But it will prob get clearer where we want to keep it as add more method interfaces so we can just change it then.

hamdiallam:
> a separate package - since it's not exactly "config". But it will prob get clearer where we want to keep it as add more me

okay sounds good! yeah i think it's useful for now to reduce the number of packages but we can rip it out again as more methods get added

PR #53, opened 2024-07-17
feat(genesis): generate genesis files

jakim929:
high level, this follows the high level approach taken in the monorepo to generate the bedrock-devnet

- uses the same deploy scripts used in the monorepo (but the monorepo script does not support multiple chains)
- creates a new python script to generate genesis for multiple l2s (follows approach in https://github.com/ethereum-optimism/optimism/blob/develop/bedrock-devnet/devnet/__init__.py)

reasons for creating a new python script is
- the monorepo version is a dependency for CI
- monorepo may or many not keep this script around in the long run

steps to generate `just generate-genesis`:
1. fetches the monorepo into a gitignored directory (opted against submodules since they are a lot more finnicky to use)
2. generate deploy-configs
3. run Deploy.s.sol to get the state dump (l1-allocs.json) and addresses (addresses.json) for each L2 chain
4. generate a single combined-l1-allocs.json file (simply merges the multiple l1s)
5. generate l2_allocs by running L2Genesis.s.sol for each L2 chain
6. generate l1_genesis.json by running op-node genesis l1 process
7. generate l2_genesis.json for each chain by running op-node genesis l2 process

once these are generated, we should commit them to the repo 

## Summary by CodeRabbit

- **New Features**
  - Introduced new Solidity smart contract `CombineAllocs` for processing allocation paths.
  - Added extensive configuration files for blockchain deployments, including addresses and deploy configs.

- **Tools and Scripts**
  - Added Python script to generate genesis configurations for L1 and L2 chains, including deployment configs, addresses, and genesis files.

- **Configuration**
  - Updated `.gitignore` to exclude `lib` directory.
  - Added new targets in `Justfile` for cleaning directories, initializing repositories, and generating genesis files.
  - Introduced new configuration option `fs_permissions` for specifying directory permissions.


jakim929:
* **#53** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/53?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

contracts/script/CombineAllocs.s.sol
```diff
@@ -0,0 +1,15 @@
...
+contract CombineAllocs is Script {
+    function setUp() public {}
+
+    function run(string[] memory allocsPaths, string memory outputPath) public {
+        for (uint256 i = 0; i < allocsPaths.length; i++) {
+            vm.loadAllocs(allocsPaths[i]);
+        }
+        vm.dumpState(outputPath);
+    }
+}
```
jakim929, line 15: this is the approach suggested in https://github.com/ethereum-optimism/optimism/pull/10777#issuecomment-2159022947 for combining multiple l1 deployments

scripts/generate-genesis.py
```diff
@@ -0,0 +1,284 @@
...
+    monorepo_default_deploy_config_template_path=monorepo_default_deploy_config_template_path,
+  )
+
+  l1_chain_id = 900
+
+  def get_l2_chain(i): 
+    return L2(l1_chain_id, 901 + i, deployer_addresses[i], supersim_dir)
+
+  l1_chain = L1(l1_chain_id, supersim_dir)
+  l2_chains = list(map(get_l2_chain, range(10)))
```
nitaliano, line 103: do you think we can reduce this down to 3-5 just so it's less json files we need to check in?

scripts/generate-genesis.py
```diff
@@ -0,0 +1,284 @@
...
+    'anvil', '--silent', '--init', l1_chain.l1_genesis_path,
+  ], cwd=paths.supersim_dir)
+  
+  wait_up(8545)
+  wait_for_rpc_server('127.0.0.1:8545')
+
+  for l2_chain in l2_chains:
+    run_command([
+        'go', 'run', 'cmd/main.go', 'genesis', 'l2',
+        '--l1-rpc', 'http://127.0.0.1:8545',
```
nitaliano, line 214: could be good to get the l1 rpc url or port into a variable so we can modify it at the top of the file if it ever needs to change

Justfile
```diff
@@ -17,4 +17,21 @@ test-go:
...
+    cd lib/optimism && \
+    git init && \
+    git remote add origin https://github.com/ethereum-optimism/optimism.git && \
+    git fetch --depth=1 origin 9001eef4784dc2950d0bdcda29752cb2939bae2b && \
+    git reset --hard FETCH_HEAD && \
+    git submodule update --init --recursive && \
+    make cannon-prestate
+
+generate-genesis: checkout-optimism-monorepo
+    python3 scripts/generate-genesis.py
```
nitaliano, line 37: It doesn't need to be in this PR more of something for us to think about. This will require everyone to also have `make` and `python3` setup already which we all do but should consider in the long run if we can get away with pulling in the monorepo in a docker container generating the files in the container and outputting them to our host fs. Would also love to get this out of python at some point but get why we're doing it this why since the monorepo does it this way.

scripts/generate-genesis.py
```diff
@@ -0,0 +1,284 @@
...
+  generate_l2_allocs(paths, l2_chains)
+  generate_l1_genesis(paths, l1_chain, l2_chains)
+  generate_l2_genesis(paths, l1_chain, l2_chains)
+  
+def generate_deploy_configs(paths, l2_chains: list[L2]):
+  for l2_chain in l2_chains:
+    # Create deploy configs that will be used by the foundry deploy script
+    deploy_config = read_json(paths.monorepo_default_deploy_config_template_path)
+    deploy_config['l1ChainID'] = l2_chain.l1_chain_id
+    deploy_config['l2ChainID'] = l2_chain.l2_chain_id
```
nitaliano, line 117: will the chain ids we setup here conflict with whats happening in our go code https://github.com/ethereum-optimism/supersim/blob/6073ce78ff834b4a400691b674bb16de5139a277/config/chain.go#L22-L40 or i'm assuming this is something we'll figure out once to start to consume these files

nitaliano (approved):

Justfile
fainashalts: Yeah I agree. We should reduce dependencies that folks need to download wherever possible. Could be good to make a follow-on issue for this.

scripts/generate-genesis.py
jakim929: yep! the next PR will use the generated files as the source of truth

Justfile
jakim929: Yep! agreed that there are way better approaches. Docker approach makes sense to me.

I don't think we should do much in the near term because
- the config/genesis generation process is currently in flux
- we're going to consolidate the process with some of the code in superchain registry
- this will be run once every few months by a single dev, then committed.

scripts/generate-genesis.py
jakim929: yep! will bring down to 5. it will also bloat our binary size

scripts/generate-genesis.py
jakim929: fixed!

Issue #17, opened 2024-06-25
op-simulator: intercept `eth_sendRawTransaction`


fainashalts:
Initially, we need to build intercept functionality that identifies the RPC calls we want to intercept and does so before passing them to the Anvil instances.

**Definition of Done**
`eth_sendRawTransaction` is intercepted by `op-simulator` based on the RPC method being called and is then sent along to the required Anvil instance.

PR #54, opened 2024-07-17
feat: add rpc intercept to `opsimulator`

tremarkley:
Closes https://github.com/ethereum-optimism/supersim/issues/17

Creates a handler in `opsimulator` that intercepts JSON RPC requests and checks which method is being called. This sets us up to be able to add special handing for interop transactions when the `eth_sendRawTransaction` method is called.

This does not handle batch transactions, we will need to update the handler to support them: https://github.com/ethereum-optimism/supersim/issues/55 

## Summary by CodeRabbit

- **New Features**
  - Added support for handling JSON-RPC requests in the simulator.
  
- **Future Enhancements**
  - Introduced a placeholder for future interoperation validation logic.

### DIFF
--- a/opsimulator/opsimulator.go
+++ b/opsimulator/opsimulator.go
@@ -1,9 +1,12 @@
 package opsimulator
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httputil"
@@ -37,6 +40,13 @@ type OpSimulator struct {
 	cfg *Config
 }
 
+type JSONRPCRequest struct {
+	Method  string        `json:"method"`
+	Params  []interface{} `json:"params"`
+	ID      int           `json:"id"`
+	JSONRPC string        `json:"jsonrpc"`
+}
+
 func New(log log.Logger, cfg *Config, l1Chain chainapi.Chain, l2Chain chainapi.Chain) *OpSimulator {
 	return &OpSimulator{
 		log:     log,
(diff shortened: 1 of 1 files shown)


tremarkley:
* **#54** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/54?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jakim929:
wdyt about using https://github.com/ethereum/go-ethereum/blob/v1.13.15/rpc/handler.go?

otherwise we have to write our own logic to handle batches, formatting errors, etc.

opsimulator/opsimulator.go
```diff
@@ -90,6 +100,37 @@ func (a *OpSimulator) Stopped() bool {
...
+
+		var req JSONRPCRequest
+		err = json.Unmarshal(body, &req)
+		if err != nil {
+			http.Error(w, "Failed to parse JSON-RPC request", http.StatusBadRequest)
+			return
+		}
+
+		if req.Method == "eth_sendRawTransaction" {
+			checkInteropInvariants()
```
jakim929, line 120: nit: can be in a separate PR, we should add the early return that write the result/response to ResponseWriter so it's easier to see how it would be used

jakim929 (approved):
looks good! ignore my comment below, I think we can add more robust json-rpc handling (with batching) later

one nit

opsimulator/opsimulator.go
tremarkley: let's do that in a separate pr. I'll have to look more into what that would look like and I want to get this in before EOD, so it is not blocking.

PR #62, opened 2024-07-25
use superchain registry AddressList as canonical type

hamdiallam:
The registry address list does not have an info about the implementation addresses which is a plus


jakim929 (approved):
looks great!

PR #75, opened 2024-07-31
Update reverse proxy to handle options requests

nitaliano:
**Description**

Updates the reverse proxy to return a `204` along with the cors headers, when the browser triggers a preflight request


fainashalts (approved):
LGTM!

PR #76, opened 2024-07-31
fix: opsimulator json parsing logic / feat: parse batch requests

hamdiallam:
Add support to parse either batch requests or normal json rpc requests.

The jsonrpc field can either be a string or numeric value. This causes 400s when a string since the current struct hard codes it to a uint. Fix this by switching this field to a `json.RawMessage` to defer decoding. We dont care about this field anyways.

### DIFF
--- a/opsimulator/opsimulator.go
+++ b/opsimulator/opsimulator.go
@@ -3,7 +3,6 @@ package opsimulator
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


opsimulator/opsimulator.go
```diff
@@ -189,28 +177,26 @@ func handler(proxy *httputil.ReverseProxy) http.HandlerFunc {
...
 			return
 		}
 
 		// TODO(https://github.com/ethereum-optimism/supersim/issues/55): support batch txs
 
-		if req.Method == "eth_sendRawTransaction" {
-			checkInteropInvariants()
+		for _, msg := range msgs {
+			if msg.Method == "eth_sendRawTransaction" {
+				checkInteropInvariants()
```
tremarkley, line 196: if one of the messages in the batch contains a tx that does not pass the interop invariant check, should we error out the entire batch or just that tx?

tremarkley (approved):

opsimulator/opsimulator.go
hamdiallam: we can error just that tx. each individual rpc within the batch has an `Error` field that we can populate

tremarkley: does this get tricky since we are also forwarding to a reverse proxy (anvil)? for example, we would need to stop txs that fail the interop checks from being forwarded to the reverse proxy, and then combine the failed txs with the response for the txs that were sent to the reverse proxy.

hamdiallam: we could filter out the failed ones when forwarding to interop but in the response populate the error if that makes sense

PR #84, opened 2024-08-05
feat: fork config rpc url override

hamdiallam:
the public endpoints in the superchain registry are rate limited endpoints.

Allow the user to override the rpc url used when forking a network setting the `SUPERSIM_RPC_URL_<name>` env variable for the chain


README.md
```diff
@@ -46,10 +46,14 @@ OPTIONS:
...
-                chains to fork in the superchain, mainnet options: [base, lyra, metal, mode, op, orderly, pgn, zora]
+                chains to fork in the superchain, mainnet options: [base, lyra, metal, mode, op,
+                orderly, pgn, superlumio, zora]. In order to replace the public rpc endpoint for
+                a chain, specify the ($SUPERSIM_RPC_URL_<CHAIN>) env variable. i.e SUPERSIM_RPC_URL_OP=http://optimism-mainnet.infura.io/v3/<API-KEY>
 
           --network value                     (default: "mainnet")               ($SUPERSIM_NETWORK)
-                superchain network. options: mainnet, sepolia, sepolia-dev-0
+                superchain network. options: mainnet, sepolia, sepolia-dev-0. In order to
+                replace the public rpc endpoint for the network, specify the
+                ($SUPERSIM_RPC_URL_<NETWORK>) env variable. i.e SUPERSIM_RPC_URL_MAINNET=http://mainnet.infura.io/v3/<API-KEY>
```
tremarkley, line 56: what is the difference between the chain rpc url (`$SUPERSIM_RPC_URL_<CHAIN>`) and the network url (`$SUPERSIM_RPC_URL_<NETWORK>`)? shouldnt there just be one rpc url that combines both chain + network?

config/cli.go
```diff
@@ -51,14 +51,14 @@ func ForkCLIFlags(envPrefix string) []cli.Flag {
...
-			Usage:    fmt.Sprintf("chains to fork in the superchain, mainnet options: [%s]", mainnetMembers),
+			Usage:    fmt.Sprintf("chains to fork in the superchain, mainnet options: [%s]. In order to replace the public rpc endpoint for a chain, specify the ($%s_RPC_URL_<CHAIN>) env variable. i.e SUPERSIM_RPC_URL_OP=http://optimism-mainnet.infura.io/v3/<API-KEY>", mainnetMembers, envPrefix),
 			Required: true,
 			EnvVars:  opservice.PrefixEnvVar(envPrefix, "CHAINS"),
 		},
 		&cli.StringFlag{
 			Name:    NetworkFlagName,
 			Value:   "mainnet",
-			Usage:   fmt.Sprintf("superchain network. options: %s", networks),
+			Usage:   fmt.Sprintf("superchain network. options: %s. In order to replace the public rpc endpoint for the network, specify the ($%s_RPC_URL_<NETWORK>) env variable. i.e SUPERSIM_RPC_URL_MAINNET=http://mainnet.infura.io/v3/<API-KEY>", networks, envPrefix),
```
tremarkley, line 61: is it possible to have an rpc url for an entire network? I thought it needs to be a combo of chain + network.

tremarkley (approved):

README.md
hamdiallam: the former is for the L2 chains. The latter is for the L1 chain

config/cli.go
hamdiallam: network here is L1, mainnet/sepolia

README.md
tremarkley: oh ok, I was kind of confused by that because you can also have an l2 "network". do you think that could be confusing? should we update to using l1 chain here instead of network?

Issue #69, opened 2024-07-29
interop contracts for forked networks


hamdiallam:
When forking, the interop predeploy contracts do not exist. We'll need to manually apply them on startup so the contracts are available

Ideas:
- Look into anvil_setCode or something at runtime to manually apply the bytecode. consider the constructor argument. If so, some storage slots to override if needed.
- Look into genesis generation and pulling the bytecode from there
- Set the dependency set information accordingly

hamdiallam:
cc @jakim929 

Perhaps by default we dont apply these changes to forked networks. Instead we have sets of experimental flags that patch remote network state `supersim fork --experiment.interop`.

PR #87, opened 2024-08-07
feat: support for adding interop contracts to forked networks

tremarkley:
Closes https://github.com/ethereum-optimism/supersim/issues/69


tremarkley:
* **#87** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/87?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

opsimulator/opsimulator.go
```diff
@@ -143,12 +143,13 @@ func (opSim *OpSimulator) Stopped() bool {
 	return opSim.stopped.Load()
 }
 
-func (opSim *OpSimulator) startStartupTasks() {
+func (opSim *OpSimulator) addDependencies() error {
 	for _, chainID := range opSim.L2Config.DependencySet {
-		opSim.startupTasks.Go(func() error {
```
tremarkley, line 148: if the dependency set is > 1, then this fails due to nonce issues. I did not see a way to set a nonce on a deposit tx, so I modified this to be a synchronous for loop.

jakim929: makes sense! we can fix later https://github.com/ethereum-optimism/supersim/issues/99

opsimulator/opsimulator.go
```diff
@@ -118,9 +113,14 @@ func (opSim *OpSimulator) Start(ctx context.Context) error {
 		}
 	}
 
-	opSim.startStartupTasks()
-	if err := opSim.startupTasks.Wait(); err != nil {
-		return fmt.Errorf("failed to start opsimulator: %w", err)
```
jakim929, line 123: nit: wdyt about putting the for loop inside a single startupTask? The main point of the startupTasks field was to better separate tasks that are run before the opsim returns ready vs. ones that can run after opsim returns ready.

That way in a separate entrypoint, a caller can just wait on startupTasks to see if opsim is ready

```
opSim.startupTasks.Go(func() error {
    for chain in ...
        addDependencySet
}
```

orchestrator/fork.go
```diff
@@ -70,19 +70,36 @@ func NetworkConfigFromForkCLIConfig(log log.Logger, envPrefix string, forkConfig
...
 			},
-		})
+		}
+
+		if l2ChainConfig.ForkConfig.UseInterop {
+			var dependencySet []uint64
+			for _, chainToAddToDepSet := range forkConfig.Chains {
+				chainToAddToDepSetCfg := config.OPChainByName(superchain, chainToAddToDepSet)
+				if chainToAddToDepSetCfg == nil {
+					return networkConfig, fmt.Errorf("unrecoginized chain %s. superchain %s", chainToAddToDepSet, superchain.Superchain)
```
jakim929, line 93: nit: typo unrecoginized

genesis/applier.go
```diff
@@ -1,27 +1,144 @@
...
-func ApplyChainID(genesisJson []byte, chainId *big.Int) ([]byte, error) {
-	var genesis core.Genesis
-	if err := json.Unmarshal(genesisJson, &genesis); err != nil {
-		return nil, fmt.Errorf("unable to parse genesis json: %w", err)
+// Uses same logic as `predeployToCodeNamespace` in Predeploy.sol to determine impl address.
+func predeployToCodeNamespace(addr common.Address) common.Address {
+	addrBigInt := new(big.Int).SetBytes(addr.Bytes())
+
+	namespaceBigInt := new(big.Int).SetBytes(common.FromHex("0xc0D3C0d3C0d3C0D3c0d3C0d3c0D3C0d3c0d30000"))
+	resultBigInt := new(big.Int).Or(new(big.Int).And(addrBigInt, big.NewInt(0xffff)), namespaceBigInt)
```
jakim929, line 50: nit: might make sense to just hard code the impl addresses as well. since there's only 3, and this is temporary code

genesis/applier.go
```diff
@@ -1,27 +1,144 @@
...
+
+	return common.BytesToAddress(resultBigInt.Bytes())
+}
+
+func fetchAllocForAddr(addr common.Address, allocsJSON []byte) (*alloc, error) {
+	var allocs allocs
+	err := json.Unmarshal(allocsJSON, &allocs)
```
jakim929, line 57: we can prob memoize the result of the unmarshal and store in memory so we don't have to unmarshal everytime. esp since this is called at the startup we should try to reduce latency as much possible

jakim929 (approved):
looks good! some minor suggestions

we should add tests to sanity check that the contracts work, but can be in separate PR

genesis/applier.go
tremarkley: I think this function is useful if we ever need to add more contracts or if the proxy addresses are ever changed for some reason. Is there a reason you prefer to hardcode the values?

opsimulator/opsimulator.go
tremarkley: sounds good, I updated the startupTasks to include `configureInteropForChain` and `addDependencies`

genesis/applier.go
tremarkley: good call, memoized it at the top level call `opsimulator.configureInteropForChain`

Issue #66, opened 2024-07-29
Supersim Release Process


hamdiallam:
- v0.0.1 release stage 0
- v0.1.0 release for stage 1
- once stable no major issues, we can do a v1.0.0 release.

Distribution
- Homebrew (`brew install supersim`)
- MacOS/Windows Go binaries in the Github release (github action)
- Build instructions in the README

PR #98, opened 2024-08-12
ci: setup release on new tags

jakim929:
context: https://github.com/ethereum-optimism/supersim/issues/66

sets up go binary build + release using goreleaser
- only way to trigger is by creating a new tag
- in the future we can figure out nightly releases / automated CI releases

to test locally, 

1. install [goreleaser](https://goreleaser.com/install/) ie. `brew install goreleaser`
2. run a release - output should be in the `dist` folder

```
 goreleaser release --snapshot --clean
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - "*"
+
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '^1.22'
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          distribution: goreleaser
+          version: "~> v2"
+          args: release --clean
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,31 @@
+version: 2
+
+before:
+  hooks:
+    - go mod tidy
+
+builds:
+  - env:
+      - CGO_ENABLED=0
+    main: ./cmd/main.go
+    goos:
+      - linux
+      - windows
+      - darwin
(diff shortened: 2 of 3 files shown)


jakim929:
* **#98** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/98?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

.goreleaser.yaml
```diff
@@ -0,0 +1,33 @@
+version: 2
```
jakim929, line 1: based on the default .goreleaser.yaml

.goreleaser.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - amd64
+      - arm64
+
+archives:
+  - format: tar.gz
+    name_template: >-
+      {{ .ProjectName }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "386" }}i386
```
nitaliano, line 25: do we need this statement for `i386`? if i'm understanding this correctly we're only building for `amd64` and `arm64`

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - "*"
```
nitaliano, line 6: we don't need to figure this out in this PR something for us to think about...it might be better if we can be more descriptive for tags so we can easily distinguish between snapshots vs full releases just by the tag name

.goreleaser.yaml
```diff
@@ -0,0 +1,33 @@
...
+      {{- else }}{{ .Arch }}{{ end }}
+      {{- if .Arm }}v{{ .Arm }}{{ end }}
+    format_overrides:
+      - goos: windows
+        format: zip
+
+changelog:
```
nitaliano, line 32: I'm wondering if we actually do want to generate changelogs by commit messages or just not have them at all for snapshot releases. I personally feel most changelogs aren't that useful based on just commit messages unless we're all following a specific pattern for it to parse.

nitaliano (approved):

.goreleaser.yaml
jakim929: ah this was copy pasta will remove

.goreleaser.yaml
jakim929: will remove changelogs for now!

Issue #102, opened 2024-08-14
Investigate how to synchronize clocks between anvil nodes


tremarkley:
The startup for each anvil node varies and because of this the timestamps on the blocks can vary by over 10 seconds between nodes. This causes issues with the timestamp invariant check on the CrossL2Inbox: https://github.com/ethereum-optimism/optimism/blob/db21f4aa016c536e4602b7b2f5effa4ac8cf35ba/packages/contracts-bedrock/src/L2/CrossL2Inbox.sol#L143 since the timestamp on the source chain can end up being over 10 seconds ahead of the timestamp on the destination chain. This seems especially prevalent when starting in fork mode.

tremarkley:
The `latestL2HeightFromL1Header` function could potentially be contributing to the clock drift in fork mode. This is run before anvil is started and involves rpcs to fetch the latest block times. I think that this could then be causing fork block numbers to not be in sync between nodes. When I remove this from the anvil startup, there is still some clock drift between nodes but it is <2 seconds.

Issue #100, opened 2024-08-13
Flaky supersim tests


tremarkley:
This issue is for investigating a root cause / fix.

hamdiallam:
what kind of errors are you seeing?

tremarkley:
The one I have noticed is the `TestStartup` test
```
=== RUN   TestStartup
    orchestrator.go:55:         INFO [08-13|17:35:42.892] starting orchestrator
    orchestrator_test.go:48: 
        	Error Trace:	/home/runner/work/supersim/supersim/orchestrator/orchestrator_test.go:48
        	Error:      	Not equal: 
        	            	expected: 0x386
        	            	actual  : 0x385
        	Test:       	TestStartup
    orchestrator.go:80:         INFO [08-13|17:35:43.224] stopping orchestrator
    anvil.go:157:               ERROR[08-13|17:35:43.226] anvil terminated with an error           role=anvil name=OPChainA chain.id=901 error="signal: killed"
    anvil.go:157:               ERROR[08-13|17:35:43.227] anvil terminated with an error           role=anvil name=OPChainB chain.id=902 error="signal: killed"
    anvil.go:157:               ERROR[08-13|17:35:43.228] anvil terminated with an error           role=anvil name=L1 chain.id=900 error="signal: killed"
--- FAIL: TestStartup (0.34s)
FAIL
FAIL	github.com/ethereum-optimism/supersim/orchestrator	0.361s
FAIL
error: Recipe `test-go` failed on line 17 with exit code 1
```

hamdiallam:
hmm that's failing on the chain id comparison which is hardcoded. Sounds like a hidden but legitimate bug

tremarkley:
I noticed that it checks the chain chain id at index 0. does the order matter or should we just check that the array contains the chain id?

hamdiallam:
ahh yea maybe the order here is messed up. Esp since the error is showing off-by-one.

Hmm since we're checking the config of the chain as well, maybe we just sort the list in the test prior to checking?

PR #103, opened 2024-08-14
fix: sync clocks across anvil nodes on startup

tremarkley:
Closes: https://github.com/ethereum-optimism/supersim/issues/102
Closes: https://github.com/ethereum-optimism/supersim/issues/100

This PR synchronizes the anvil block timestamps across all L2 nodes prior to supersim startup. This is done using the following steps:
1. start the nodes in automine mode so no new blocks are created
2. turn on 2 second interval mining for all nodes using a goroutine after all anvil nodes have marked themselves as ready using `evm_setIntervalMining`

This synchronization is not perfect but it should ensure that the clocks between anvil nodes never drift further than 2 seconds apart.

### DIFF
--- a/config/chain.go
+++ b/config/chain.go
@@ -24,39 +24,6 @@ var (
 		Mnemonic:       "test test test test test test test test test test test junk",
 		DerivationPath: accounts.DefaultRootDerivationPath,
 	}
-
-	DefaultNetworkConfig = NetworkConfig{
-		L1Config: ChainConfig{
-			Name:          "L1",
-			ChainID:       genesis.GeneratedGenesisDeployment.L1.ChainID,
-			SecretsConfig: DefaultSecretsConfig,
-			GenesisJSON:   genesis.GeneratedGenesisDeployment.L1.GenesisJSON,
-		},
-		L2Configs: []ChainConfig{
-			{
-				Name:          "OPChainA",
-				ChainID:       genesis.GeneratedGenesisDeployment.L2s[0].ChainID,
-				SecretsConfig: DefaultSecretsConfig,
-				GenesisJSON:   genesis.GeneratedGenesisDeployment.L2s[0].GenesisJSON,
-				L2Config: &L2Config{
-					L1ChainID:     genesis.GeneratedGenesisDeployment.L1.ChainID,
-					L1Addresses:   genesis.GeneratedGenesisDeployment.L2s[0].RegistryAddressList(),
-					DependencySet: []uint64{genesis.GeneratedGenesisDeployment.L2s[1].ChainID},
-				},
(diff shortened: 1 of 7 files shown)


tremarkley:
* **#103** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/103?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jakim929 (changes requested):
I like the approach to start mining at the same time! Especially in fork mode where the time to start mining depends on the speed of the RPC url being used (since anvil does an initial fetch).

Not 100% sure, but I think the chains start with the correct same timestamps, but all start mining (or being active) at different times?

I think we can simplify this / reduce amount of async calls made to anvil post startup by
- (already implemented) start all anvil initially without `--block-time 2`
- guarantee that all chains start out with the same timestamp
	- For fork mode, it seems like the `latestL2HeightFromL1Header` logic correctly returns the L2 heights that has the same timestamp, so we prob won't need to change anything here. Anvil should correctly fetch the l2 block to start on, and they should all have the same (or very similar) timestamps.
	- For vanilla mode, we can pass a `--timestamp` flag to anvil with the exact timestamp we want them to start on - should be same across all anvil instances. Anvil ignores `--timestamp` in fork mode but honors it in vanilla mode
- once all anvils are active, we call `evm_setIntervalMining` on them. If we run the calls in parallelized go routines there should be minimal difference between each chain.

I think we can avoid calling `evm_setNextBlockTimestamp` or `eth_getBlockByNumber` (HeaderByNumber), since we can guarantee the that the chains are started initially with the same timestamp.

hamdiallam:
> For fork mode, it seems like the latestL2HeightFromL1Header logic correctly returns the L2 heights that has the same timestamp, so we prob won't need to change anything here. Anvil should correctly fetch the l2 block to start on, and they should all have the same (or very similar) timestamps.

This is roughly true. It looks for an L2 height which should approximately be in the same range as all the other chains.

Agree with @jakim929 thought. Maybe on `eth_sendRawTransaction` we could also progress the timestamp of a chain that hasn't had any txs yet?

config/chain.go
```diff
@@ -90,6 +57,9 @@ type ChainConfig struct {
 
 	// Optional Config (L1 chain if nil)
 	L2Config *L2Config
+
+	// Optional
+	StartingTimestamp int64
```
jakim929, line 62: should be uint64

jakim929 (approved):
looks great!

config/chain.go
tremarkley: good catch!

PR #109, opened 2024-08-15
Adds Top Level CORS MIddleware

nitaliano:
**Description**

Recently pulled the latest version of supersim and was running into cors issues again. We recently updated how the reverse proxy is created and how we forward the requests. The issue is we aren't adding the cors header back to the response. The last issue was related to not handling options request.

This PR adds a top level cors middleware that handles options requests and applies the cors headers to all responses so we can avoid breaking this again in the future.


hamdiallam (approved):

hamdiallam:
looks good to me. A lot cleaner to use middleware

PR #119, opened 2024-08-22
Update README.md

jakim929:
adds addresses for CrossL2Inbox and L2ToL2CrossDomainMessenger


fainashalts (approved):
LGTM! We should expose this in the CLI somehow I think.

PR #124, opened 2024-08-29
fix(opsim): improve startup time

jakim929:
Noticed startup time was very slow these days. This slows down iteration and tests.
- Most of the startup time was waiting for opsim to get ready
- Mostly spent waiting for the tx receipt on the tx to set the dependencySet
- Since we run anvil in automine mode in the beginning, the tx should land very fast but still asynchronously (with a small delay AFTER the rpc call returns, since anvil internally has a job queue for mining)
- updated logic for checking tx to check at faster tick intervals since we assume the tx will land. The default bind.WaitMined checks every second (leading to it taking 2 seconds)

Before
Screenshot 2024-08-29 at 4 35 42 PM

After 
Screenshot 2024-08-29 at 4 45 03 PM

### DIFF
--- a/opsimulator/opsimulator.go
+++ b/opsimulator/opsimulator.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 	"sync/atomic"
+	"time"
 
 	ophttp "github.com/ethereum-optimism/optimism/op-service/httputil"
 	"github.com/ethereum-optimism/optimism/op-service/predeploys"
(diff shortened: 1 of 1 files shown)


jakim929:
* **#124** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/124?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

hamdiallam (approved):

PR #140, opened 2024-09-12
feat: interop logging & code cleanup

hamdiallam:
Add a logging section for interop when enabled

```
(EXPERIMENTAL) Interop!
-----------------------
For more information see the explainer! ( https://docs.optimism.io/stack/protocol/interop/explainer )

Added Predeploy Contracts:
 - L2ToL2CrossDomainMessenger: 0x4200000000000000000000000000000000000007
 - CrossL2Inbox:               0x4200000000000000000000000000000000000022
 ```

Code Cleanup
- move applier logic to the interop, called by the orchestrator and not individually by op-sim
- no need to setup task groups since the application happens synchronously in a loop
- update config so the interop enabled can be seen at the network level and not at the chain level

### CODE REVIEW

**Comment by hamdiallam** [comment]
* **#140** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @hamdiallam and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

File: orchestrator/orchestrator.go [hunk 1]
```diff
@@ -148,37 +165,15 @@ func (o *Orchestrator) Stop(ctx context.Context) error {
...
-
-	handleErr(o.l1Chain.SetIntervalMining(ctx, nil, 2))
-	wg.Done()
-
 	for _, chain := range o.l2Chains {
-		go func() {
-			defer wg.Done()
-			handleErr(chain.SetIntervalMining(ctx, nil, 2))
-		}()
+		if err := chain.SetIntervalMining(ctx, nil, 2); err != nil {
```
tremarkley, line 172: previously when this was in a go routine we could kick off mining on each chain in parallel, this reduces clock drift and also speeds up the startup time. is there a reason you prefer this to be synchronous?

opsimulator/opsimulator.go
```diff
@@ -155,24 +134,6 @@ func (opSim *OpSimulator) EthClient() *ethclient.Client {
 	return opSim.ethClient
 }
 
-func (opSim *OpSimulator) startStartupTasks(ctx context.Context) {
-	opSim.startupTasks.Go(func() error {
```
tremarkley, line 159: does removing this go routine slowdown startup time since now we will not be configuring interop for each chain in parallel?

config/cli.go
```diff
@@ -71,10 +71,10 @@ func ForkCLIFlags(envPrefix string) []cli.Flag {
 			EnvVars: opservice.PrefixEnvVar(envPrefix, "NETWORK"),
 		},
 		&cli.BoolFlag{
-			Name:    InteropEnabledInForkModeFlagName,
-			Value:   false,
-			Usage:   "Enable interop in fork mode",
-			EnvVars: opservice.PrefixEnvVar(envPrefix, "FORK_WITH_INTEROP"),
+			Name:    InteropEnabledFlagName,
+			Value:   true, // enabled by default
```
tremarkley, line 75: I might be remembering incorrectly, but I thought we decided that for fork mode interop would not be enabled by default.

opsimulator/opsimulator.go
hamdiallam: per my earlier comment they weren't being init'd in parallel

config/cli.go
hamdiallam: hmm yea i'm actually not sure, @jakim929 do you remember?


with interop being a big focus now, may be fine to enable by default, idk

orchestrator/orchestrator.go
hamdiallam: ahh wait i'm seeing `go func() { .. }` within that for loop now. Let me revert that piece. So it's concurrent

Let me revert this

hamdiallam: okay reverted to make this concurrent, good catch. Missed that

opsimulator/opsimulator.go
hamdiallam: Let me also make this concurrent in the orchestrator

hamdiallam: okay pushed, startup time is back to being fast again

jakim929 (approved):
lgtm!

PR #146, opened 2024-09-16
example: add CrossChainPingPong.sol

jakim929:
an example contract that wraps L2CDM to use in guides.

contract is deployed on multiple chains, and allows hitting the "ball" back and forth between the chains using the L2CDM

### DIFF
--- a/contracts/src/CrossChainPingPong.sol
+++ b/contracts/src/CrossChainPingPong.sol
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.25;
+
+import {IL2ToL2CrossDomainMessenger} from "@contracts-bedrock/L2/interfaces/IL2ToL2CrossDomainMessenger.sol";
+import {Predeploys} from "@contracts-bedrock/libraries/Predeploys.sol";
+
+error CallerNotL2ToL2CrossDomainMessenger();
+
+error InvalidCrossDomainSender();
+
+struct PingPongBall {
+    uint256 rallyCount;
+    uint256 lastHitterChainId;
+    address lastHitterAddress;
+}
+
+contract CrossChainPingPong {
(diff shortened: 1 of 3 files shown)


jakim929:
* **#146** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/146?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

contracts/src/CrossChainPingPong.sol
```diff
@@ -0,0 +1,96 @@
...
+        emit BallReceived(IL2ToL2CrossDomainMessenger(MESSENGER).crossDomainMessageSource(), _ball);
+    }
+
+    function _sendBallMessage(PingPongBall memory _ball, uint256 _toChainId) internal {
+        bytes memory _message = abi.encodeCall(this.receiveBall, (_ball));
+        IL2ToL2CrossDomainMessenger(MESSENGER).sendMessage(_toChainId, address(this), _message);
+    }
+
+    function _isBallOnThisChain() internal view returns (bool) {
+        return receivedBall.lastHitterAddress != address(0);
```
tremarkley, line 90: if checking for the ball on the chain, why does this check the lastHitterAddress? Is it just meant to check that receivedBall is defined so this check could really be for any of the fields on `receivedBall`? nit but it might be good to add a comment for why this check is done this way, since I am not sure how intuitive that is from just reading this one line of code and not having an understanding of the entire contract.

tremarkley (approved):
this is cool!

contracts/src/CrossChainPingPong.sol
jakim929: yeah I should've added more comments - this is just a hacky way of checking the receivedBall is defined - will add a comment!

PR #156, opened 2024-09-19
fix: minor improvement to README

smartcontracts:
Just makes that first tutorial thing a little bit clearer


jakim929 (approved):
thanks for the update!

Issue #135, opened 2024-09-09
Support interoperable WETH in Supersim


fainashalts:
Like we did for SuperchainERC20 we should add support for interop WETH by predeploying the contracts and creating a sample flow for how it could be used.

**Definition of Done**

- [x] WETH contracts are predeployed
- [ ] L2CDM wrapper contract for wrapping / unwrapping ETH
- [ ] Documentation includes examples of how to interact with them/use them

**Resources**
https://github.com/ethereum-optimism/specs/blob/main/specs/interop/superchain-weth.md
https://github.com/ethereum-optimism/optimism/pull/11042

PR #167, opened 2024-09-20
feat: `SuperchainETHWrapper` contract

tremarkley:
Closes https://github.com/ethereum-optimism/supersim/issues/135

This contract facilitates sending ETH across chains within the Superchain by wrapping ETH into `SuperchainWETH`, relaying the `SuperchainWETH` to the destination chain using `L2ToL2CrossDomainMessenger`, and then unwrapping it back to ETH and sending it to the specified recipient on the destination chain. This is not meant to be used in production and should only be used in testnet or for prototypes.


tremarkley:
* **#167** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/167?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

zainbacchus:
OH LETS GOOOOOO

interop/relayer.go
```diff
@@ -119,16 +117,11 @@ func (r *L2ToL2MessageRelayer) Start(indexer *L2ToL2MessageIndexer, clients map[
 					}
 					// Pad gas by 33%.
 					paddedGas := (gasEstimate / 3) + gasEstimate
-					mu.Lock()
```
tremarkley, line 122: this mutate does not work properly when multiple messages are sent in the same transaction. simplest way to fix this is to remove the lock and move the transactor within the case block

tremarkley:
Once https://github.com/ethereum-optimism/optimism/pull/12514 merges, then I will update this to include latest changes to `SuperchainWETH`

contracts/src/SuperchainETHWrapper.sol
```diff
@@ -0,0 +1,86 @@
...
+ * @dev The contract integrates with the SuperchainWETH contract for wrapping and unwrapping ETH,
+ *      and uses the L2ToL2CrossDomainMessenger for relaying the wrapped ETH between chains.
+ */
+contract SuperchainETHWrapper {
+    /**
+     * @dev Emitted when ETH is received by the contract.
+     * @param from The address that sent ETH.
+     * @param value The amount of ETH received.
+     */
+    event LogReceived(address from, uint256 value);
```
nitaliano, line 39: might be overkill but having the `from` address indexed might be helpful for debugging

nitaliano (approved):

contracts/src/SuperchainETHWrapper.sol
hamdiallam: rename to `EthReceived()`?

hamdiallam (approved):

contracts/src/SuperchainETHWrapper.sol
tremarkley: done

Issue #170, opened 2024-09-23
L2 to L2 Message Passing appears broken


fainashalts:
**Environment information**
 - Operating System: macOS Sonoma 14.5
 - Supersim Version: 0.1.0-alpha.9

**Foundry version.**
forge 0.2.0 (ea3ba89 2024-09-05T00:22:21.218185000Z)

**Describe the bug**
`cast send` L2 <> L2 ERC20 transfer appears to be broken. 

**Steps to reproduce**

1. Follow the steps [here](https://supersim.pages.dev/getting-started/first-steps) for "Send an interoperable SuperchainERC20 token from chain 901 to 902 (L2 to L2 message passing)"

**Observed behavior**: `cast balance --erc20 0x0bEa8920a4FfB1888Ec3Ac1BC0D23f414B0a28cA 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --rpc-url http://127.0.0.1:9546` yields 0

**Expected behavior**: balance should be 1000.

PR #172, opened 2024-09-24
fix(relayer): force private key curve to make tx send correctly

jakim929:
closes: https://github.com/ethereum-optimism/supersim/issues/170

## bug:
when building the binary with CGO_ENABLED=0, relayer fails to send tx with `private key curve is not secp256k1`. Only repro-able when building using that flag (hence our releases are affected)

Geth checks the curve with an equality check, - similar issue in the monorepo can be found here https://github.com/ethereum-optimism/optimism/pull/11382/files#diff-4b062d5d864b6672e6c935962b0e1ad449ee0fe6365a0b770b3b3b7c535f0f2b

## fix:
manually update the curve to the correct instance. if this becomes a more commonly used pattern to send tx, we can create some helper functions around it until the issue is solved

## notes:
- also updates to use the new `devkeys` package in the monorepo, so deleted `hdaccounts` pacakge.


jakim929:
* **#172** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/172?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tremarkley (approved):

PR #190, opened 2024-10-04
update unit tests to read from vars instead of secrets

nitaliano:
**Description**

Updates our unit tests to use public endpoints instead of private ones


nitaliano:
This has the run for it https://github.com/ethereum-optimism/supersim/actions/runs/11184654740

jakim929 (approved):

PR #196, opened 2024-10-09
feat: autogenerate golang bindings

jakim929:
`generate-bindings.sh` script will
1. Downloads a tar.gz file containing contract artifacts from a specified URL.
2. Extracts ABIs for specified contracts from the artifacts.
3. Uses `abigen` to generate Go bindings for each contract.
4. Outputs the generated bindings to a specified directory.

`just generate-monorepo-bindings` should generate and output the bindings into the bindings module

abigen is not smart enough yet to deduplicate same-named variables ie. `ICrossL2InboxIdentifier` which is generated in both L2ToL2CDM and CrossL2Inbox so there may be some manual work to be done for now


jakim929:
* **#196** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/196?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Justfile
```diff
@@ -38,5 +38,8 @@ calculate-artifact-url:
     checksum=$(bash scripts/ops/calculate-checksum.sh) && \
     echo "https://storage.googleapis.com/oplabs-contract-artifacts/artifacts-v1-$checksum.tar.gz"
 
+generate-monorepo-bindings:
```
tremarkley, line 41: should we include this with generate-genesis? it seems like we would always want to run them together

tremarkley (approved):
great job!

Justfile
jakim929: Didn't think it was too impactful yet since we don't have to run both in CI right now.

I can add a `generate-all` command that runs both though!

tremarkley: yeah, that sounds good! I just think it's a potential footgun for a dev to forget to run one but not the other

PR #203, opened 2024-10-17
Improve Dockerfile

janjakubnanista:
**Description**

- Leverage multi-stage build in the `Dockerfile` to speed up the docker build by enabling the individual stages to run in parallel
  - Add `foundry` stage that downloads `foundry`
  - Add `builder` stage that builds the project
  - Add `runner` stage that combines the binaries from the two other stages
- Parameterize the base image using the `BASE_IMAGE` build argument
- Used `ENTRYPOINT` instead of `CMD`. This allows user to simply `docker run` the image to execute `supersim` with no arguments or, if they need to, pass `supersim` arguments directly to `docker run`
- Added `.dockerignore` to exclude metafiles & docs from being copied to the builder image

There are several more improvements possible (e.g. `go` is not really required in the `runner` stage) but have not been added for the sake of brevity

**Tests**

- Add `RUN` commands to the `Dockerfile` to ensure that the required binaries exist

**Additional context**

**Metadata**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,15 @@
 # by specifying BASE_IMAGE build argument
 ARG BASE_IMAGE=golang:1.22
 
+#   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-
+#  / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \
+# `-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'
 # 
-# This stage builds the foundry binaries
+#           This stage builds the foundry binaries
 # 
+#   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-
+#  / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \
+# `-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'
 FROM $BASE_IMAGE AS foundry
 
 # Make sure foundryup is available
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,19 +1,54 @@
-FROM golang:1.22
+# The base image for the supersim image can be modified
+# by specifying BASE_IMAGE build argument
+ARG BASE_IMAGE=golang:1.22
 
-RUN apt-get update && apt-get install -y curl git
+# 
+# This stage builds the foundry binaries
+#
```
nitaliano, line 7: for the comments on the different stages it might be better if we add comments like this https://github.com/ethereum-optimism/ecosystem/blob/ef2845eaecf99581be17c2020fb53e05096e94da/Dockerfile#L10-L12 so it's easier to see the different stages at a glance

janjakubnanista: @nitaliano In my dockerfiles I usually go even crazier, just wasn't sure whether you'd like it. [Here's an example](https://github.com/LayerZero-Labs/devtools/blob/main/Dockerfile#L35), what do you think?

```yaml
#   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-
#  / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \
# `-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'
#
#          This stage builds the foundry binaries
#
#   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-
#  / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \
# `-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'
```

janjakubnanista:
@nitaliano I changed the comment style and made two more changes - `ENTRYPOINT` and `.dockerignore`. I personally like to test everything I can so maybe in another PR I could add a workflow to test the docker image. 

Another thing that would be nice is to have a workflow that builds & pushes the image to GHCR/Dockerhub.

Dockerfile
nitaliano: haha nice! yea totally that works as well just something that helps the developer understand how many stages exist at a glance

nitaliano:
@janjakubnanista agree great suggestion we are planning on adding docker building/publishing with docker using circleci so it hits labs GCP account for publishing public images

janjakubnanista:
@nitaliano Have you considered github workflows? I have used both and recently prefer to keep everything in one place. The two are almost feature-equal and the github workflows setup creates less friction and more transparency IMO.

Also I the tests failing on `Proxy: implementation not initialized` which seems unrelated to the changes I made - is it a flaky test or should I investigate?

janjakubnanista:
@nitaliano here's an example of very complex distributed mutli-platform docker image build & push pipeline I made for LayerZero's devtools just to demonstrate the setup https://github.com/LayerZero-Labs/devtools/blob/platforms/.github/workflows/reusable-publish-docker.yaml

nitaliano:
@janjakubnanista for the unit tests I think you just need to have it rebased with this https://github.com/ethereum-optimism/supersim/commit/8debcc6abce6f55b8e5b228eadde58a035a13b01 we can merge it in without that though

we prefer to use circleci across the whole `ethereum-optimism` org when it comes to dealing with secrets, i don't disagree we like github actions much better

nitaliano:
Thanks again for the PR!

PR #206, opened 2024-10-17
Add docker image tests

janjakubnanista:
**Description**

- Adding smoke tests for `linux/amd64` and `linux/arm64` images to the CI pipeline. These tests check that the `Dockerfile` can be built as well as that the `supersim` executable runs fine when the image is run

**Tests**

- Yesssss

**Additional context**

**Metadata**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -118,7 +118,7 @@ jobs:
         uses: docker/build-push-action@v6
         with:
           platforms: ${{ matrix.platform }}
-          outputs: type=docker,supersim:dev
+          outputs: type=docker,optimism/supersim:dev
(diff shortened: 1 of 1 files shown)


nitaliano:
just getting back to this @janjakubnanista i'm wondering if something is up with the `linux/arm64` tests these are taking a very long time to finish compared to the `amd64` one

janjakubnanista:
@nitaliano ah yeah the emulation is quite computationally intensive (github machines are amd64 and need to use QEMU when running docker for other architectures). They can be removed since the only other solution is to use a different runner which is a paid thing.

nitaliano:
yea i think this is good to merge in once we get rid of `arm64` once this gets merged in would like to make the docker test as a required CI check

PR #210, opened 2024-10-18
CI pins go version to whats in go.mod

nitaliano:
**Description**

I noticed that `golangci-lint` is [failing to install ](https://github.com/ethereum-optimism/supersim/actions/runs/11408059484/job/31745440552#step:3:38) on a couple of PRs now when we're pulling in go `1.23` so this PR just pins to the go version in our go.mod file to avoid this. We'll most likely need to bump our `golangci-lint` version once we want to update our go version

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -10,7 +10,7 @@ runs:
     - name: Setup Go
       uses: actions/setup-go@v4
       with:
-        go-version: '^1.22'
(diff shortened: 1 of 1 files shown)


nitaliano:
`TestForkedInteropInvariantCheckSucceeds` keeps failing with the `gasPrice` error but other than that this is good to go

hamdiallam (approved):

PR #221, opened 2024-10-23
feat(monorepo): use forge install to install monorepo

jakim929:
as per @hamdiallam 's suggestion just using forge install to make it easier to install.

Previously we were using scripts from the monorepo but now we're only using contracts (to generate artifacts url0, so moving to inside the contracts folder

Tested on an empty repo that the commands work correctly.

for later PR
- see if we can remove the monorepo-commit-hash file
- will add a script to make sure the go.mod version for the go package doesn't drift from the contracts version


jakim929:
* **#221** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/221?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

hamdiallam (approved):
thanks for handling this!

PR #236, opened 2024-10-28
[ISSUE-227] add `docs` CLI command

### LINKED ISSUE #227: Add CLI command to display interop spec & optimism docs in Supersim flow
When a user runs `supersim --docs` they should get a list of the following clickable links:

[Superchain Interop Specs](https://specs.optimism.io/interop/overview.html)
[Optimism Documentation](https://docs.optimism.io/)
[Supersim Documentation](https://supersim.pages.dev/)

jameskim0987:
**Description**
Implemented `docs` CLI command for displaying the 3 available documentation page as clickable links.

**Tests**
Manual testing. Output:

```
$ ./main docs
Here are the available documentation links:

Superchain Interop Specs
Optimism Documentation
Supersim Documentation
```

**Additional context**
N/A

**Metadata**
Resolves #227


jameskim0987:
Hi @fainashalts, this PR is ready for review.

I wasn't sure about the exact output format or the output strings so I kept them minimal to start with. Let me know if it needs to be reformatted/reworded!

config/cli.go
```diff
@@ -128,6 +137,11 @@ func ReadCLIConfig(ctx *cli.Context) (*CLIConfig, error) {
 		LogsDirectory: ctx.String(LogsDirectoryFlagName),
 	}
 
+	if ctx.Bool(DocsFlagName) {
+		printDocLinks()
+		os.Exit(0)
+	}
+
```
jakim929, line 144: instead of exiting here, it would be better to return the flag inside the cfg, then cancel in main.go#SupersimMain using `closeApp` if the flag is set. That aligns better with using `closeApp` as the main way to shutdown.

jakim929 (changes requested):
Thank you for your contribution! Mostly looks good - a few suggested changes.

Also, high level, wdyt about making this a command instead of flag? `supersim docs` instead of `supersim --docs`. Think it might be better as a command but will leave that up to you

config/cli.go
jakim929: if u end up making this a command, you can check in main `ctx.Command` before running `closeApp`

config/cli.go
```diff
@@ -167,3 +181,20 @@ func (c *CLIConfig) Check() error {
 
 	return nil
 }
+
+func printDocLinks() {
+	links := []struct {
```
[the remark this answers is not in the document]
jameskim0987: Done!

config/cli.go
jameskim0987: Done!

jameskim0987:
> Thank you for your contribution! Mostly looks good - a few suggested changes.
> 
> Also, high level, wdyt about making this a command instead of flag? `supersim docs` instead of `supersim --docs`. Think it might be better as a command but will leave that up to you

Thank you for the suggestions! I reflected your suggested change from `docs` flag to a command as it makes more sense to me as well.

Regarding exit steps after printing the links -- in addition to invoking `closeApp`, I have followed up with `os.Exit(0)` since without the exit call the app outputs supersim logs which is not necessary for the `docs` command.

```
$ ./main docs
Here are the available documentation links:

Superchain Interop Specs
Optimism Documentation
Supersim Documentation
INFO [11-01|19:32:20.788] starting supersim
WARN [11-01|19:32:20.791] err piping stderr to log file            role=anvil name=Local chain.id=900 err="read |0: file already closed"
t=2024-11-01T19:32:20-0400 lvl=crit msg="Application Failed" err="failed to start: orchestrator failed to start: l1 chain Local failed to start: failed to start anvil: context canceled\ncontext canceled"
```

Let me know how it looks!

jakim929 (approved):

Issue #144, opened 2024-09-13
refactor SuperchainERC20 event logging in opsim


jakim929:
L2NativeSuperchainERC20 and SuperchainWETH logging can use the same code. We'll prob have more token types soon too.

s4rv4d:
Can I work on this ? Is anyone taking this up ?

tabshaikh:
@jakim929 made a pr for the changes

s4rv4d:
@jakim929 @fainashalts can you pls clarify is this issue is still open, i saw there are merged related to SuperchainERC20 event logging on opsimulator, and L2NativeSuperchainERC20 and SuperchainWETH use ICrosschainERC20 for common event emits so is there work required here and what is the scope?

tabshaikh:
@s4rv4d had renamed the pr incorrectly but the pr makes a common logging function for both `L2NativeSuperchainERC20` and `SuperchainWETH` which use `ICrosschainERC20`. so i think this issue is complete

tabshaikh:
@jakim929 can be closed :)

PR #253, opened 2024-11-08
💅 refactor L2NativeSuperchainERC20 and SuperchainWETH logging

tabshaikh:
**Description**

updated L2NativeSuperchainERC20 and SuperchainWETH to use similar logging

**Metadata**

https://github.com/ethereum-optimism/supersim/issues/144

### DIFF
--- a/admin/admin_test.go
+++ b/admin/admin_test.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"testing"
+	"time"
 
 	"github.com/ethereum-optimism/optimism/op-service/testlog"
 	"github.com/ethereum/go-ethereum/log"
@@ -32,6 +33,9 @@ func TestAdminServerBasicFunctionality(t *testing.T) {
 	require.Equal(t, "OK", string(body))
 
 	require.NoError(t, adminServer.Stop(context.Background()))
+
+	// Add a small delay to ensure the server has fully stopped
+	time.Sleep(100 * time.Millisecond)
 
 	resp, err = http.Get(fmt.Sprintf("%s/ready", adminServer.Endpoint()))
 	if err == nil {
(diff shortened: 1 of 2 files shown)


jakim929 (changes requested):
I should have specified it better in the issue, but I think a cleaner approach is to use the SuperchainERC20 binding for both contracts since they share the same interface (bindings/superchainerc20.go). 

Roughly something like
```go
opSim.bgTasks.Go(func() error {
    contracts := []struct {
        name    string
        address common.Address
    }{
        {"SuperchainWETH", predeploys.SuperchainWETHAddr},
        {"L2NativeSuperchainERC20", common.HexToAddress(l2NativeSuperchainERC20Addr)},
    }

    for _, c := range contracts {
        contract, err := bindings.NewSuperchainERC20(c.address, opSim.Chain.EthClient())
        if err != nil {
            return fmt.Errorf("failed to create %s contract: %w", c.name, err)
        }

        mintEventChan := make(chan *bindings.SuperchainERC20CrosschainMint)
        burnEventChan := make(chan *bindings.SuperchainERC20CrosschainBurn)

        mintSub, err := contract.WatchCrosschainMint(&bind.WatchOpts{Context: opSim.bgTasksCtx}, mintEventChan, nil)
        if err != nil {
            return fmt.Errorf("failed to subscribe to %s#CrosschainMint: %w", c.name, err)
        }

        burnSub, err := contract.WatchCrosschainBurn(&bind.WatchOpts{Context: opSim.bgTasksCtx}, burnEventChan, nil)
        if err != nil {
            return fmt.Errorf("failed to subscribe to %s#CrosschainBurn: %w", c.name, err)
        }
...
```

Please let me know if I can provide any more context

tabshaikh:
@jakim929 pushed changes using `SuperchainERC20`. Was previously abstracting a lot because thought there were no common interface but as they both use `ICrosschainERC20` things are much cleaner

jakim929 (approved):

jakim929:
@tabshaikh thanks this looks great! good to go once CI passes

tabshaikh:
weirdly the admin test was failing fixed that
Screenshot 2024-11-09 at 1 40 28 AM

PR #264, opened 2024-11-13
feat: Completed implementation of customizable anvil hosts

### LINKED ISSUE #262: Allow customizing anvil hosts
context: https://github.com/ethereum-optimism/supersim/issues/250

- add a string CLI flag `--host`
- by default this should be unset or set to `127.0.0.1`
- add a `Host` parameter to ChainConfig
- anvil.go should pass the host specified in the config when starting anvil
- admin server should also use the same host
- add unit tests to make sure this works correctly

s29papi:
**Description**

Adding customizable host configuration options for anvil nodes. This allows users to specify custom host addresses for L1 and L2 instances.

✅ Implementation Complete

Features:
- Added L1Host and L2Host configuration flags
- Implemented host configuration in chain and network configs
- Added host validation and default fallbacks
- Integrated with anvil startup and orchestrator

Tests verify:
- Default host fallback (127.0.0.1)
- Custom host configurations (0.0.0.0)
- Mixed host scenarios
- Host validation
- Integration with anvil and orchestrator

**Additional context**

This implementation allows users to run anvil instances on different host addresses, enabling:
- Custom network configurations
- Flexible deployment scenarios
- Better container support
- Improved network isolation options

**Metadata**

Closes #262


s29papi:
@jakim929 The customizable host implementation is now complete! 

You can now configure host addresses using the CLI flags

orchestrator/orchestrator_test.go
```diff
@@ -20,45 +17,121 @@ type TestSuite struct {
...
 
-func createTestSuite(t *testing.T) *TestSuite {
+type TestCase struct {
+	name    string
+	l1Host  string
+	l2Host  string
+	wantErr bool
+}
+
+func createTestSuite(t *testing.T, tc *TestCase) *TestSuite {
```
jakim929, line 27: think it will be more flexible to just pass it a networkConfig as a param

```
func createTestSuite(t *testing.T, networkConfig *NetworkConfig) *TestSuite {
```

orchestrator/orchestrator_test.go
```diff
@@ -20,45 +17,121 @@ type TestSuite struct {
...
+		l1Host = tc.l1Host
+		l2Host = tc.l2Host
+	}
+
+	networkConfig.L1Config.Host = l1Host
+	networkConfig.L1Config.Port = 9471
+	for i := range networkConfig.L2Configs {
+		networkConfig.L2Configs[i].Host = l2Host
+		networkConfig.L2StartingPort = 9473
+	}
```
jakim929, line 43: can we leave the starting ports to remain default 0? The test run in parallel so it will clash on port usage.

orchestrator/orchestrator_test.go
```diff
@@ -20,45 +17,121 @@ type TestSuite struct {
...
+			t.Logf("Testing with L1 Host: %s, L2 Host: %s", tc.l1Host, tc.l2Host)
+
+			testSuite := createTestSuite(t, &tc)
+			if !tc.wantErr {
+				require.NotNil(t, testSuite)
+				verifyChainConfigs(t, testSuite)
+			}
+		})
+	}
+}
```
jakim929, line 113: the createTestSuite should ideally not have any tests in it.

If you want to test for the orchestrator failing to start I would add another test that runs

```
	orchestrator, err := NewOrchestrator(testlog, closeApp, &networkConfig)
```

and checks for error with out using createTestSuite

orchestrator/orchestrator_test.go
```diff
@@ -20,45 +17,121 @@ type TestSuite struct {
...
 }
 
-func createTestSuite(t *testing.T) *TestSuite {
+type TestCase struct {
+	name    string
+	l1Host  string
+	l2Host  string
+	wantErr bool
+}
+
```
jakim929, line 26: to remove - see comment below - these can just be individual tests that don't use createTestSuite

orchestrator/orchestrator.go
```diff
@@ -49,7 +51,10 @@ func NewOrchestrator(log log.Logger, closeApp context.CancelCauseFunc, networkCo
 	// Sping up OpSim to fornt the L2 instances
 	for i := range networkConfig.L2Configs {
 		cfg := networkConfig.L2Configs[i]
-		l2OpSims[cfg.ChainID] = opsimulator.New(log, closeApp, nextL2Port, l1Anvil, l2Anvils[cfg.ChainID], l2Anvils, networkConfig.InteropDelay)
+		if cfg.Host == "" {
+			cfg.Host = defaultHost
+		}
```
jakim929, line 56: by default the cli flags will fall back to 127.0.0.1 so this can be removed, no need to have `defaultHost`

anvil/anvil.go
```diff
@@ -36,7 +36,7 @@ var (
 )
 
 const (
-	host                 = "127.0.0.1"
+	defaultHost          = "127.0.0.1"
```
jakim929, line 39: this can be removed, since host will default to "127.0.0.1" based on the default CLI flag value

orchestrator/orchestrator.go
```diff
@@ -31,6 +31,8 @@ type Orchestrator struct {
 	l2ToL2MsgRelayer *interop.L2ToL2MessageRelayer
 }
 
+const defaultHost = "127.0.0.1"
+
```
jakim929, line 35: to remove

orchestrator/orchestrator.go
jakim929, line 56: right now the CLI flags are not passed to the L2 chains. I think you need to pass something like

```
cfg.Host = networkConfig.L2Host
```

orchestrator/orchestrator.go
```diff
@@ -49,7 +51,10 @@ func NewOrchestrator(log log.Logger, closeApp context.CancelCauseFunc, networkCo
 	// Sping up OpSim to fornt the L2 instances
 	for i := range networkConfig.L2Configs {
 		cfg := networkConfig.L2Configs[i]
-		l2OpSims[cfg.ChainID] = opsimulator.New(log, closeApp, nextL2Port, l1Anvil, l2Anvils[cfg.ChainID], l2Anvils, networkConfig.InteropDelay)
+		if cfg.Host == "" {
```
jakim929, line 54: right now the CLI flags are not passed to the chains for the L1 chain as well. I think you need to pass something like

cfg.Host = networkConfig.L1Host

opsimulator/opsimulator.go
```diff
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	host                        = "127.0.0.1"
+	defaultHost                 = "127.0.0.1"
```
jakim929, line 36: remove default host, make it always passed down

jakim929 (changes requested):
Thank you for the contribution! This looks pretty close - just some changes requested to how the config is passed down from the CLI flags, and also some test changes

orchestrator/orchestrator.go
s29papi: Hey man thanks for the reviews. 

Based on my understanding of this, I see the 

```	// Forward set ports. Setting `0` will work to allocate a random port
	networkConfig.L1Config.Port = cliConfig.L1Port
	networkConfig.L2StartingPort = cliConfig.L2StartingPort```
	
```	// Forward interop config
	networkConfig.InteropAutoRelay = cliConfig.InteropAutoRelay
	networkConfig.InteropDelay = cliConfig.InteropDelay```
	
cliConfig params are forwaded to the network config from outside the orchestrator so i decided to follow same 

``` // Forward host config
	networkConfig.L1Config.Host = cliConfig.L1Host
	for i := range networkConfig.L2Configs {
		networkConfig.L2Configs[i].Host = cliConfig.L2Host
	}```
	
But I might be wrong, My thinking is that the l1Chain is handled by this 

```	// Spin up L1 anvil instance
	l1Anvil := anvil.New(log, closeApp, &networkConfig.L1Config)
```

s29papi:
@jakim929 

Changes:
1. Removed default hosts
2. Removed checks in createsuite
3. Seperates host tests on orchestrator

Would appreciate if you clarify the comment "cli flags are not passed down", they are currently being forwarded from supersim and also the test suite, and printing to the console within anvil, and opsim also suggests this.

jakim929 (approved):
This looks correct and apologize for the delay.

by cli flag not being passed down I just meant that we don't need a default value to be compared against at every level - the implementation looks correct now

anvil/anvil_test.go
```diff
@@ -15,7 +15,21 @@ import (
 )
 
 func TestAnvil(t *testing.T) {
-	cfg := config.ChainConfig{ChainID: 10, Port: 0}
+	t.Run("unset default host", func(t *testing.T) {
+		cfg := config.ChainConfig{ChainID: 10, Port: 0}
+		testAnvilInstance(t, cfg)
+	})
```
jakim929, line 21: this run should fail because anvil correctly expects a host.

can delete this test

Issue #204, opened 2024-10-17
supersim admin API


jakim929:
**general**
- JSON RPC so it's easily usable with `cast rpc`

**some example endpoints**
- ready(): used for checking whether the service is ready to use
- getConfig(): returns config of the chains (RPC URLs, addresses, etc.
- getL1Addresses(chainId): returns contract addresses for given chainId
- getL2Addresses(chainId): same as above but for L2 contracts
- getL2ToL2MessageByMsgHash(msgHash)
- getL2ToL2MessageByTxHash(msgHash)
- getL1ToL2MessageByTxHash(l1DepositTxHash)
- listL1ToL2Messages(optional chainId)
- listL2ToL2Messages(optional chainId)

PR #265, opened 2024-11-14
feat(admin): added rpc method for getL2ToL2 Message(msgHash)

s4rv4d:
**Description**
added getL2ToL2 Message(msgHash), essentially letting devs fetch L2ToL2Message via msgHash

**Metadata**

- adds extra functionality requested in #204

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -3,6 +3,7 @@ package admin
 import (
 	"context"
 	"fmt"
+	"math/big"
 	"net"
 	"net/http"
 	"sync"
@@ -10,6 +11,7 @@ import (
 	"github.com/ethereum-optimism/supersim/config"
 	"github.com/ethereum-optimism/supersim/interop"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/gin-gonic/gin"
@@ -37,6 +39,15 @@ type RPCMethods struct {
 type JSONRPCError struct {
 	Code    int    `json:"code"`
 	Message string `json:"message"`
+}
+
+type JSONL2ToL2Message struct {
+	Destination uint64         `json:"Destination"`
+	Source      uint64         `json:"Source"`
+	Nonce       *big.Int       `json:"Nonce"`
+	Sender      common.Address `json:"Sender"`
+	Target      common.Address `json:"Target"`
+	Message     hexutil.Bytes  `json:"Message"`
 }
 
 func (e *JSONRPCError) Error() string {
@@ -176,7 +187,7 @@ func (m *RPCMethods) GetL1Addresses(args *uint64) (*map[string]string, error) {
 	return &reply, nil
 }
 
-func (m *RPCMethods) GetL2ToL2MessageByMsgHash(args *common.Hash) (*interop.L2ToL2Message, error) {
+func (m *RPCMethods) GetL2ToL2MessageByMsgHash(args *common.Hash) (*JSONL2ToL2Message, error) {
 
 	if m.l2ToL2MsgIndexer == nil {
 		return nil, &JSONRPCError{
@@ -199,6 +210,16 @@ func (m *RPCMethods) GetL2ToL2MessageByMsgHash(args *common.Hash) (*interop.L2To
 			Message: fmt.Sprintf("Failed to get message: %v", err),
 		}
 	}
+
+	msg := storeEntry.Message()
+
 	m.log.Debug("admin_getL2ToL2MessageByMsgHash")
-	return storeEntry.Message(), nil
+	return &JSONL2ToL2Message{
+		Destination: msg.Destination,
+		Source:      msg.Source,
+		Nonce:       msg.Nonce,
+		Sender:      msg.Sender,
+		Target:      msg.Target,
+		Message:     msg.Message,
+	}, nil
 }
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -34,6 +34,19 @@ type RPCMethods struct {
 	l2ToL2MsgIndexer *interop.L2ToL2MessageIndexer
 }
 
+type JSONRPCError struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+}
+
+func (e *JSONRPCError) Error() string {
+	return e.Message
+}
+
+func (err *JSONRPCError) ErrorCode() int {
+	return err.Code
+}
+
 func NewAdminServer(log log.Logger, port uint64, networkConfig *config.NetworkConfig, indexer *interop.L2ToL2MessageIndexer) *AdminServer {
 
 	adminServer := &AdminServer{log: log, port: port, networkConfig: networkConfig}
@@ -113,12 +126,9 @@ func (s *AdminServer) setupRouter() *gin.Engine {
 
 	rpcServer := rpc.NewServer()
 	rpcMethods := &RPCMethods{
-		log:           s.log,
-		networkConfig: s.networkConfig,
-	}
-
-	if s.networkConfig.InteropEnabled {
-		rpcMethods.l2ToL2MsgIndexer = s.l2ToL2MsgIndexer
+		log:              s.log,
+		networkConfig:    s.networkConfig,
+		l2ToL2MsgIndexer: s.l2ToL2MsgIndexer,
 	}
 
 	if err := rpcServer.RegisterName("admin", rpcMethods); err != nil {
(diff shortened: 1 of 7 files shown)


admin/admin.go
```diff
@@ -145,6 +160,32 @@ func (m *RPCMethods) GetL1Addresses(args *uint64) *map[string]string {
...
+		}
+
+		reply := storeEntry.Message()
+
+		return reply
+
+	} else {
+		m.log.Error("auto relay not enabled", "args", *args)
+		return nil
+	}
```
jakim929, line 190: the indexer runs regardless of the auto relayer. maybe some thing like "Interop is not enabled, please enable interop with --interop.enabled"

admin/admin.go
```diff
@@ -145,6 +160,32 @@ func (m *RPCMethods) GetL1Addresses(args *uint64) *map[string]string {
...
+		storeEntry, err := m.l2ToL2MsgIndexer.Get(*args)
+
+		if err != nil {
+			m.log.Error("failed to get msg", "error", err)
+			return nil
+		}
+
+		reply := storeEntry.Message()
+
+		return reply
```
jakim929, line 185: nit 
```
return storeEntry.Message()
```

admin/admin_test.go
```diff
@@ -23,7 +23,7 @@ func TestAdminServerBasicFunctionality(t *testing.T) {
 	testlog := testlog.Logger(t, log.LevelInfo)
```
jakim929, line 24: can u add a test to see that the result is returned correctly? or alternatively can be in supersim_test

admin/admin.go
```diff
@@ -20,18 +22,27 @@ type AdminServer struct {
...
 
-func NewAdminServer(log log.Logger, port uint64, networkConfig *config.NetworkConfig) *AdminServer {
-	return &AdminServer{log: log, port: port, networkConfig: networkConfig}
+func NewAdminServer(log log.Logger, port uint64, networkConfig *config.NetworkConfig, indexer *interop.L2ToL2MessageIndexer) *AdminServer {
+
+	adminServer := &AdminServer{log: log, port: port, networkConfig: networkConfig}
+
+	if networkConfig.InteropEnabled && indexer != nil {
+		adminServer.l2ToL2MsgIndexer = indexer
+	}
```
jakim929, line 43: nit- if it's nil then it will be initialized correctly
```
	adminServer := &AdminServer{log: log, port: port, networkConfig: networkConfig, l2ToL2MsgIndexer: indexer}
```

admin/admin.go
```diff
@@ -102,8 +113,12 @@ func (s *AdminServer) setupRouter() *gin.Engine {
...
 	rpcServer := rpc.NewServer()
 	rpcMethods := &RPCMethods{
-		Log:           s.log,
-		NetworkConfig: s.networkConfig,
+		log:           s.log,
+		networkConfig: s.networkConfig,
+	}
+
+	if s.networkConfig.InteropEnabled {
+		rpcMethods.l2ToL2MsgIndexer = s.l2ToL2MsgIndexer
```
jakim929, line 121:
```
rpcMethods.l2ToL2MsgIndexer = s.l2ToL2MsgIndexer
```

no need to check for interopEnabled. from the perspective of the admin server, if it gets passed a non-nil indexer it assumes it's usable

admin/admin.go
jakim929: also this should not be a log it should return an RPC error with the correct code

supersim.go
```diff
@@ -66,9 +66,9 @@ func NewSupersim(log log.Logger, envPrefix string, closeApp context.CancelCauseF
 		return nil, fmt.Errorf("failed to create orchestrator")
 	}
 
-	adminServer := admin.NewAdminServer(log, cliConfig.AdminPort, &networkConfig)
+	a := admin.NewAdminServer(log, cliConfig.AdminPort, &networkConfig, o.L2ToL2MessageIndexer())
```
jakim929, line 69: since now there's a dependency on the orchestrator's internal properties, we should move the admin server start / end logc into orchestrator

orchestrator/orchestrator.go
```diff
@@ -211,6 +211,10 @@ func (o *Orchestrator) L2Chains() []config.Chain {
 	return chains
 }
 
+func (o *Orchestrator) L2ToL2MessageIndexer() *interop.L2ToL2MessageIndexer {
+	return o.l2ToL2MsgIndexer
+}
+
```
jakim929, line 217: see comment on supersim.go

once the admin server start logic is moved into here, we can remove this fn

admin/admin.go
jakim929, line 185: can u verify that this returns correctly (serialized and formatted correctly) when u call the json rpc server? and add a test for it?

admin/admin.go
```diff
@@ -145,6 +160,32 @@ func (m *RPCMethods) GetL1Addresses(args *uint64) *map[string]string {
...
+func (m *RPCMethods) GetL2ToL2MessageByMsgHash(args *common.Hash) *interop.L2ToL2Message {
+
+	if args == nil {
+		m.log.Error("valid msg hash not provided", "args", *args)
+		return nil
+	}
+
+	if m.networkConfig.InteropEnabled && m.l2ToL2MsgIndexer != nil {
+
+		storeEntry, err := m.l2ToL2MsgIndexer.Get(*args)
```
jakim929, line 176: also need to handle case if the storeEntry just doesn't exist - should end up returning nil

admin/admin.go
```diff
@@ -145,6 +160,32 @@ func (m *RPCMethods) GetL1Addresses(args *uint64) *map[string]string {
...
 	return &reply
 }
+
+func (m *RPCMethods) GetL2ToL2MessageByMsgHash(args *common.Hash) *interop.L2ToL2Message {
+
+	if args == nil {
+		m.log.Error("valid msg hash not provided", "args", *args)
+		return nil
+	}
+
```
jakim929, line 173: nit: early return is better than the if/else statement below

if m.l2ToL2MsgIndexer == nil {
	m.log.Error("auto relay not enabled", "args", *args)
	return nil
}

^ we should return a better error

admin/admin.go
```diff
@@ -145,6 +160,32 @@ func (m *RPCMethods) GetL1Addresses(args *uint64) *map[string]string {
...
+	m.log.Debug("admin_getL2Addresses")
 	return &reply
 }
+
+func (m *RPCMethods) GetL2ToL2MessageByMsgHash(args *common.Hash) *interop.L2ToL2Message {
+
+	if args == nil {
+		m.log.Error("valid msg hash not provided", "args", *args)
+		return nil
+	}
```
jakim929, line 172: can we validate that it's a correct msgHash?

jakim929 (changes requested):
thanks for the PR - some changes requested

admin/admin.go
s4rv4d: did to a test on local


will add test cases too

admin/admin_test.go
s4rv4d: added to supersim_test

s4rv4d:
adding a screenshot of test results on local

Screenshot 2024-11-17 at 4 40 58 pm

s4rv4d:
Updated rpc response

`{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "Destination": 902,
    "Source": 901,
    "Nonce": 0,
    "Sender": "0x4200000000000000000000000000000000000028",
    "Target": "0x4200000000000000000000000000000000000028",
    "Message": "0x7cfd6dbc000000000000000000000000420beef000000000000000000000000000000001000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb92266000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb9226600000000000000000000000000000000000000000000000000000000000003e8"
  }
}`

jakim929 (approved):

supersim_test.go
```diff
@@ -65,6 +65,15 @@ type TestSuite struct {
...
 }
 
+type JSONL2ToL2Message struct {
+	Destination uint64         `json:"Destination"`
+	Source      uint64         `json:"Source"`
+	Nonce       *big.Int       `json:"Nonce"`
+	Sender      common.Address `json:"Sender"`
+	Target      common.Address `json:"Target"`
+	Message     hexutil.Bytes  `json:"Message"`
+}
```
jakim929, line 75: you can prob just import this from admin.JSONL2ToL2Message

In a future PR might make sense to unify with the indexer's message type so we don't have two types. As discussed, out of scope for this PR! just noting.

jakim929 (approved):
looks good, can u rebase? will run CI checks then

supersim_test.go
s4rv4d: noted

PR #266, opened 2024-11-14
Fix tests failing in ci

tabshaikh:
**Description**
fixes tests failing in ci

### DIFF
--- a/supersim_test.go
+++ b/supersim_test.go
@@ -967,7 +967,7 @@ func TestInteropInvariantSucceedsWithDelay(t *testing.T) {
 
 func TestInteropInvariantFailsWhenDelayTimeNotPassed(t *testing.T) {
 	testSuite := createInteropTestSuite(t, config.CLIConfig{
-		InteropDelay: 5,
+		InteropDelay: 2, // 2 second delay
 	})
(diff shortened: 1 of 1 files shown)


tabshaikh:
@jakim929 made a pr to fix the go-tests failing in the main branch

supersim_test.go
```diff
@@ -967,7 +967,7 @@ func TestInteropInvariantSucceedsWithDelay(t *testing.T) {
 
 func TestInteropInvariantFailsWhenDelayTimeNotPassed(t *testing.T) {
 	testSuite := createInteropTestSuite(t, config.CLIConfig{
-		InteropDelay: 5,
+		InteropDelay: 2, // 2 second delay
```
tremarkley, line 970: can you explain more about why lowering the delay from 5 to 2 makes this test less flaky?

tabshaikh: @tremarkley the value here does not matter, just keeping it same as the previous test https://github.com/ethereum-optimism/supersim/blob/c6c140211c1477e25dc1c03c6540e2b015dc9d47/supersim_test.go#L910 . The error in tests was because of the next lines changed

tabshaikh:
think the tests have been fixed, and doesn't require this pr anymore

Issue #274, opened 2024-11-19
TicTacToe frontend uses a hardcoded address


tabshaikh:
**Describe the bug**
as the ticTacToe frontend uses a hardcoded address, there is a blank page if create2 contract address is different

fainashalts:
Thanks for raising this @tabshaikh and I see you made a PR to fix! 🤩 Looks like this is resolved so I'm going to close but if I'm missing something comment here and I'll reopen.

PR #275, opened 2024-11-19
fix: take address from cli for tictactoe frontend

supercontracts:
**Description**
The tictactoe frontend uses a hardcoded address which may fail on changes. so adding a cli flag `--address=`. 
How to run it: 
```
npm run dev --address=0x123...
# or
npm run build --address=0x123...
```

#274


examples/tictactoe/src/constants/tictactoe.ts
```diff
@@ -1,4 +1,4 @@
-export const address = '0x81D8CB015e36922F7ecC5CEAdd018dc7327e6FC7'
+export const address = import.meta.env.TICTACTOE_ADDRESS || '0x81D8CB015e36922F7ecC5CEAdd018dc7327e6FC7'
```
hamdiallam, line 1: If we're going to be supplying the address. Lets instead not have a default provided and simply resort to the cli flag

Is it possible for the process to fail if this flag isn't supplied?

hamdiallam (changes requested):
Thanks for the contribution! This direction makes sense

examples/tictactoe/src/constants/tictactoe.ts
tabshaikh: yeah its possible make a change

hamdiallam:
ty!! will locally test this and merge in after getting off my flight :)

hamdiallam (changes requested):
Can you remove the `yarn.lock` file

hamdiallam:
Looks good! Sorry one last request to also remove the `package-lock.json` file as well since we use pnpm for dependency management and not npm directly

otherwise good to go

hamdiallam:
@tabshaikh npm_config style reading doesn't work for pnpm so superseding this functionality with env vars in #288

PR #280, opened 2024-11-25
feat(admin): getL1ToL2MessageByTxHash(l1DepositTxHash)

s4rv4d:
**Description**
added getL1ToL2MessageByTxHash(l1DepositTxHash), essentially letting devs fetch L1ToL2Message via depositTxnHash

**Metadata**

- adds extra functionality requested in #204

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ethereum-optimism/supersim/config"
 	"github.com/ethereum-optimism/supersim/interop"
+	"github.com/ethereum-optimism/supersim/opsimulator"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
@@ -26,6 +27,7 @@ type AdminServer struct {
 
 	networkConfig    *config.NetworkConfig
 	l2ToL2MsgIndexer *interop.L2ToL2MessageIndexer
+	l1DepositStore   *opsimulator.L1DepositStoreManager
 
 	port uint64
 }
@@ -34,6 +36,7 @@ type RPCMethods struct {
 	log              log.Logger
 	networkConfig    *config.NetworkConfig
 	l2ToL2MsgIndexer *interop.L2ToL2MessageIndexer
+	l1DepositStore   *opsimulator.L1DepositStoreManager
 }
 
 type JSONRPCError struct {
@@ -50,6 +53,17 @@ type JSONL2ToL2Message struct {
 	Message     hexutil.Bytes  `json:"Message"`
 }
 
+type JSONDepositTx struct {
+	SourceHash          common.Hash     `json:"SourceHash"`
+	From                common.Address  `json:"From"`
+	To                  *common.Address `json:"To"`
+	Mint                *big.Int        `json:"Mint"`
+	Value               *big.Int        `json:"Value"`
+	Gas                 uint64          `json:"Gas"`
+	IsSystemTransaction bool            `json:"IsSystemTransaction"`
+	Data                hexutil.Bytes   `json:"Data"`
+}
+
 func (e *JSONRPCError) Error() string {
 	return e.Message
 }
@@ -58,9 +72,9 @@ func (err *JSONRPCError) ErrorCode() int {
 	return err.Code
 }
 
-func NewAdminServer(log log.Logger, port uint64, networkConfig *config.NetworkConfig, indexer *interop.L2ToL2MessageIndexer) *AdminServer {
+func NewAdminServer(log log.Logger, port uint64, networkConfig *config.NetworkConfig, indexer *interop.L2ToL2MessageIndexer, l1DepositStore *opsimulator.L1DepositStoreManager) *AdminServer {
 
-	adminServer := &AdminServer{log: log, port: port, networkConfig: networkConfig}
+	adminServer := &AdminServer{log: log, port: port, networkConfig: networkConfig, l1DepositStore: l1DepositStore}
 
 	if networkConfig.InteropEnabled && indexer != nil {
 		adminServer.l2ToL2MsgIndexer = indexer
@@ -140,6 +154,7 @@ func (s *AdminServer) setupRouter() *gin.Engine {
 		log:              s.log,
 		networkConfig:    s.networkConfig,
 		l2ToL2MsgIndexer: s.l2ToL2MsgIndexer,
+		l1DepositStore:   s.l1DepositStore,
 	}
 
 	if err := rpcServer.RegisterName("admin", rpcMethods); err != nil {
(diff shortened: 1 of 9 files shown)


opsimulator/deposits.go
```diff
@@ -36,7 +36,7 @@ type LogSubscriber interface {
 }
 
 // transforms Deposit event logs into DepositTx
-func SubscribeDepositTx(ctx context.Context, logSub LogSubscriber, depositContractAddr common.Address, ch chan<- *types.DepositTx) (ethereum.Subscription, error) {
+func SubscribeDepositTx(ctx context.Context, logSub LogSubscriber, depositContractAddr common.Address, ch chan<- *types.DepositTx, chLog chan<- *types.Log) (ethereum.Subscription, error) {
```
jakim929, line 39: high level it would be cleaner to just pass a single channel here. 

if you need other info from the depositTx like some of the log data, you can create a wrapper that includes the deposit tx along with the log

opsimulator/opsimulator.go
```diff
@@ -148,8 +152,9 @@ func (opSim *OpSimulator) startBackgroundTasks() {
 	// Relay deposit tx from L1 to L2
 	opSim.bgTasks.Go(func() error {
 		depositTxCh := make(chan *types.DepositTx)
+		l1DepositTxnLogCh := make(chan *types.Log)
 		portalAddress := common.Address(opSim.Config().L2Config.L1Addresses.OptimismPortalProxy)
-		sub, err := SubscribeDepositTx(context.Background(), opSim.l1Chain.EthClient(), portalAddress, depositTxCh)
+		sub, err := SubscribeDepositTx(context.Background(), opSim.l1Chain.EthClient(), portalAddress, depositTxCh, l1DepositTxnLogCh)
```
jakim929, line 157: similar to the indexer/relayer design for l2 to l2 messages, it will be more maintainable to 

have the indexer have a start function and run separately from the opsimulator. the op simulator then "listens" to the indexer to get the stream of deposit tx

opsimulator/store.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type L1DepositStore struct {
+	entryByHash map[common.Hash]*types.DepositTx
```
jakim929, line 12: some high level thoughts

1. entryByHash should be indexed by something that globally & uniquely identifies a depositTX. the l1 txhash is insufficient because 
2. you should prob maybe store more than just the deposit tx here since the indexer likely wants more info than that (ie. the log where this was emitted for instance)

opsimulator/opsimulator.go
```diff
@@ -168,6 +174,9 @@ func (opSim *OpSimulator) startBackgroundTasks() {
 				}
 
 				opSim.log.Info("OptimismPortal#depositTransaction", "l2TxHash", depTx.Hash().String())
+				if err := opSim.l1DepositStoreManager.Set(l1Log.TxHash, dep); err != nil {
```
jakim929, line 177: a single l1 transaction can have multiple deposit tx inside it so it cannot be used as the index. instead the deposit tx's own transaction hash might be a better one

s4rv4d: got it makes sense updating this

opsimulator/opsimulator.go
s4rv4d: while indexer listens for new tranction a transaction event should also be sent from indexer to opsim to SendTransaction to done while the indexer store ref to the initial depTxn sub message

s4rv4d: or another way could be both opsim and indexer can both sub to deposit txn, the op sim executes transaction while the indexer just stores ref?

opsimulator/deposits.go
s4rv4d: update: sticking to the single channel

opsimulator/store.go
s4rv4d: update:
1. using the source hash as the key value, since it uniquely identifies the source of the deposit

s4rv4d: 2. updated the message to contain both the depositTxn and the initiating log in the form of a struct

s4rv4d:
**Responsibilities**:

_Indexer_ - store/update deposit txn in store
_Opsim_ - listen to indexer and execute deposit txn

jakim929:
Sorry for the delay here. This is looking pretty good! 

A few suggestions for the high level architecture

1. A single instance of the indexer ideally should handle all chains (not 1 chain per indexer).
2. The indexer should be initialized at the orchestrator for ALL chains.
3. The indexer instance is passed to the opsim, where it listens to events for a specific chain.

I think both the admin endpoint and the opsim should query the indexer (not the store manager). 

This provides better separation of responsibility because the only one that is in charge of managing the lifecycle of the indexer, and opsim / admin only use it on a readonly basis. This is similar to the approach for the l2 to l2 indexer.

Will leave comments on specific lines for this!

opsimulator/opsimulator.go
```diff
@@ -123,6 +128,10 @@ func (opSim *OpSimulator) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to create eth client: %w", err)
 	}
 
+	if err := opSim.indexer.Start(ctx, opSim.l1Chain.EthClient(), opSim.Chain); err != nil {
+		return fmt.Errorf("L1ToL2Indexer failed to start: %w", err)
+	}
```
jakim929, line 133: Orchestrator should start the indexer so remove logic here

opsimulator/indexer.go
```diff
@@ -0,0 +1,130 @@
...
+			HandleCrit: func(err error) {
+				fmt.Printf("unhandled indexer error: %v\n", err)
+			},
+		},
+		tasksCtx:    tasksCtx,
+		tasksCancel: tasksCancel,
+	}
+}
+
+func (i *L1ToL2MessageIndexer) Start(ctx context.Context, client *ethclient.Client, l2Chain config.Chain) error {
```
jakim929, line 44: should accept an array of l2 chains - maybe this method can be come a private helper method `startForChain`?

and the new `Start` can loop through the array of chains and call this. Just a suggestion - many ways to do this

opsimulator/indexer.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	if err := i.storeManager.Set(dep.SourceHash, &depositMessage); err != nil {
+		i.log.Error("failed to store deposit tx to chain: %w", "chain.id", chainID, "err", err)
+		return err
+	}
+
+	i.eb.Publish(depositMessageInfoKey(chainID), depTx)
+	return nil
+}
```
jakim929, line 130: We should add a test at the indexer level in indexer_test.go. it should confirm that sending a tx gets correctly picked up, and the subscribe methods work correctly.

opsimulator/deposits.go
```diff
@@ -22,9 +22,20 @@ type depositTxSubscription struct {
 	doneCh          chan struct{}
 }
 
+type DepositChannels struct {
+	DepositTxCh chan<- *types.DepositTx
+	LogCh       chan<- types.Log
+}
```
jakim929, line 28: We should ideally just remove this file altogether, and in the indexer logic just subscribe to the log.

The indexer can subscribe to the log, then manually convert it to a deposit tx.

admin/admin.go
```diff
@@ -223,3 +255,56 @@ func (m *RPCMethods) GetL2ToL2MessageByMsgHash(args *common.Hash) (*JSONL2ToL2Me
...
+	}
+
+	if (args == nil || args == &common.Hash{}) {
+		return nil, &JSONRPCError{
+			Code:    -32602,
+			Message: "Valid msg hash not provided",
+		}
+	}
+
+	storeEntry, err := m.l1DepositStore.Get(*args)
```
jakim929, line 274: This should query the indexer not the store directly.

jakim929 (changes requested):
Left a few high level comments about the design, please let me know if they require clarification!

orchestrator/orchestrator.go
```diff
@@ -228,7 +249,7 @@ func (o *Orchestrator) L1Chain() config.Chain {
 
 func (o *Orchestrator) L2Chains() []config.Chain {
 	var chains []config.Chain
-	for _, chain := range o.l2OpSims {
```
s4rv4d, line 231: @jakim929  this change was in main, and was causing linting issues, so updated it to l2Chain

opsimulator/indexer_test.go
```diff
@@ -86,3 +93,62 @@ func TestSubscribeDepositTx(t *testing.T) {
...
+		log := <-logCh
+		depTx := types.NewTx(dep)
+
+		initiatedDepTxn[i] = depTx
+
+		err := indexer.ProcessEvent(dep, log, chain.Config().ChainID)
+		require.NoError(t, err, "Should send valid details")
+	}
+
+	for i := 0; i < len(initiatedDepTxn); i++ {
```
s4rv4d, line 139: @jakim929 added test case to test publish

supersim_test.go
```diff
@@ -1132,3 +1125,73 @@ func TestAdminGetL2ToL2MessageByMsgHash(t *testing.T) {
 	assert.Equal(t, tx.To().String(), message.Target.String())
 	assert.Equal(t, tx.To().String(), message.Sender.String())
 }
+
+func TestAdminGetL1ToL2MessageByTxnHash(t *testing.T) {
+	t.Parallel()
+
+	testSuite := createTestSuite(t)
```
s4rv4d, line 1132: @jakim929 updated test case over here

PR #309, opened 2024-12-21
feat: specify number of L2s to spin up

hamdiallam:
We generate genesis files for 5 possible L2s. 5 is sufficient so introduce a flag to specify
the count with a max 5, with a default of 2.

- removed orchestrator_test which doesn't test anything that supersim_test doesnt
- refactored test setup so that the same logic is used for cli config


hamdiallam:
* **#309** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/309?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/309?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

supersim_test.go
tremarkley, line 1: nit: I think it would be a good idea to add a test that tests that supersim can be started when set to the the max l2 countof 5. that would catch any issues where we didnt generate enough l2 states.

tremarkley (approved):

PR #333, opened 2025-02-11
chore: Update CircleCI configuration and GitHub token handling

raffaele-oplabs:
**Description**

This pull request includes several changes to the `.circleci/config.yml` file to update the CircleCI configuration. The most important changes involve updating the version of the `ethereum-optimism/circleci-utils` orb and modifying how the GitHub token is retrieved and used in the jobs.

Updates to CircleCI configuration:

* Updated the `ethereum-optimism/circleci-utils` orb version from `0.0.13` to `1.0.8` in the `executors` section.
* Replaced the `utils/get-github-access-token` command with a custom `run` command to set the `GITHUB_TOKEN` environment variable in the `jobs` section.
* Removed the `utils/get-github-access-token` command in another part of the `jobs` section, simplifying the configuration.

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ executors:
 
 orbs:
   go: circleci/go@2.2.3
-  utils: ethereum-optimism/circleci-utils@0.0.13
+  utils: ethereum-optimism/circleci-utils@1.0.8
 
 commands:
   # By default, CircleCI does not checkout any submodules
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -100,11 +100,13 @@ jobs:
       - checkout-with-submodules
       - install-dependencies
       - install-go-modules
-      - utils/get-github-access-token:
+      - run:
+          name: Get GitHub token
+          command: |
+            echo "export GITHUB_TOKEN=$GITHUB_TOKEN_GOVERNANCE" >> $BASH_ENV
```
nickbalestra, line 106: just to understand,  any reason on why we want to avoid using `utils/get-github-access-token` for this? has that been removed from v1 ?

.circleci/config.yml
```diff
@@ -130,7 +132,6 @@ jobs:
       - checkout
       - attach_workspace:
           at: /tmp/docs
-      - utils/get-github-access-token
```
nickbalestra, line 133: this job doesn't require the access token anymore? I see it was replace in the `go-releases` job with a different step but not in this, am I missing something?

.circleci/config.yml
raffaele-oplabs: The previous version is less secure. This upgrade removes some issues we had previously.

.circleci/config.yml
raffaele-oplabs: Not required anymore.

.circleci/config.yml
nickbalestra: Sorry, but not sure I follow - If upgrading remove some issues, why we not using the built-in utility anymore (`utils/get-github-access-token`) and instead we replace with inline code ?

nickbalestra (approved):

PR #347, opened 2025-02-21
fix: bump required anvil version

tremarkley:
There is a bug in older versions of anvil that causes supersim to break so bumping to the min version of anvil that contains the fix: https://github.com/foundry-rs/foundry/releases/tag/nightly-51b75c83045a963a48f9cd8d765e0f2e775acb3e

This bug is present when running > 2 l2 chains and causes `eth_call`'s to fail with:
```
error code -32003: Insufficient funds for gas * price + value
```

When looking into this it seems like this happens due to the way older versions of anvil handle deposit txs.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -33,7 +33,7 @@ var (
 )
 
 const (
-	minAnvilTimestamp = "2025-02-10T09:00:00.000000000Z"
+	minAnvilTimestamp = "2025-02-17T00:00:00.000000000Z"
 )
 
 func main() {


tremarkley:
* **#347** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/347?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/347?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jakim929 (approved):

jakim929 (changes requested):
Hmm this will make it hard for people to download foundry

tarunkhasnavis (approved):

tremarkley:
@jakim929 I dont think so, this bug is present on v1.0.0 of foundry

fainashalts:
@tremarkley any idea why this only happens when you have more than 2 L2 chains? Feels like handling of deposit transactions wouldn't change based on the number of L2s we're running but I totally could be missing something!

tremarkley:
@fainashalts dug into it quite a bit yesterday and not sure why when there are more than one deposit txs sent that it causes this issue. will need to look into it a bit more to determine the root cause, but for now upgrading anvil does fix the issue.

jakim929:
looked into this with @tremarkley and the issue is due to mismatch with the L1Block contract at `0x4200000000000000000000000000000000000015` between the deployed contract <> what `revm` expects

when running anvil with `--optimism` without a specified `--hardfork` it defaults to `latest`, and revm will use `isthmus`. One update on `isthmus` is an update to the storage L1Block where https://github.com/ethereum-optimism/optimism/blob/aaee96abf5ebcc0be0797597cc5cfb5de0b3c69d/packages/contracts-bedrock/src/L2/L1Block.sol#L60 where `operatorFeeConstant` will be at the same slot.

so the deployed contract has `EnumerableSet.UintSet dependencySet;` at the same slot. EnumerableSet has the array of all elements in the set in the first slot, and hence the length of the array will be what `revm` reads when it tries to read `operatorFeeConstant`, it will read `dependencySet.length` instead.

So as we add more chains to the dependency set, it will cause revm to add a constant value to the required balance for the tx. ie. with just one chain in a dependency set, every eth_call will cost 1 more wei, and with 2 chains in dependency set, every eth_call will cost 2 more wei etc. Since the balance of 0x0000...000 which is used in the gas estimation is set to 1 in the genesis file, any time we have more that 1 chain in a dep set, EVERY eth_call will fail.


### fix: 
- we can set the hardfork to `granite`  so revm doesn't assume operator fee exists
- hackier: we can set the balance of 0x000..000 to larger than the max number of chains - then we'll still be able to use Pectra features like 7702

### long term fix:
when interop ships in a future hardfork, this won't be an issue, because the L1BlockInterop will inherit from the proper L1Block. When that operator fee contract PR is merged, should be trivial to pull in those contracts into the genesis state.

### side note:
- we should add holocene and (isthmus when time comes) to https://github.com/foundry-rs/foundry/blob/c609884bdb13b9846fe9ddc5f08d99cf30c53695/crates/anvil/src/hardfork.rs#L173

tremarkley:
going to close this in favor of #348

PR #349, opened 2025-02-26
feat: support superchain-registry 2.0

jakim929:
### context

closes https://github.com/ethereum-optimism/supersim/issues/332

after https://github.com/ethereum-optimism/superchain-registry/pull/816 the superchain-registry no longer exports the chain configs / superchain.

We were previously depending on it for 
- types for chain / superchain / addresses config
- constant definitions for superchains/chains 

### changes

1. uses the types in op-geth instead https://github.com/ethereum-optimism/op-geth/blob/optimism/superchain/superchain.go
2. creates a wrapper type RegistrySuperchain and RegistryChain to capture the identifier which is the file name (different from chain.name)
3. vendors the entire folder from https://github.com/ethereum-optimism/superchain-registry/tree/main/superchain/configs
4. embeds the chain config toml files using go:embed, and at package init() stores in variable
5. adds scripts to update to latest commit hash & copy over the configs folder into `registry/vendor-superchain-registry`

### notes

- just cloning into temp directory rather than submodules because submodules is a bit more tedious & we only need a small subset of the static files
- to use go:embed, the files need to be inside the module
- `just update-and-vendor-superchain-registry` to update


jakim929:
* **#349** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/349?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/349?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

registry/init.go
tremarkley, line 1: was this file copied in from somewhere else?

tremarkley (approved):
wow, nice job!

fainashalts (approved):
Looks like we're storing all the chain `.toml` files now. Would be nice to add info somewhere on how a dev should update this when there is a new registry version, to be sure nothing is missed. Otherwise LGTM!

jakim929:
@fainashalts  

the command is `just update-and-vendor-superchain-registry` it's available in the Justfile! it's not a flow that the users of this repo need to run so we don't need to add it to README i think

registry/init.go
jakim929: there was a similar file in superchain-registry previously (no longer exists) but I adapted it to be compatible with the latest registry schema

fainashalts:
> @fainashalts
> 
> the command is `just update-and-vendor-superchain-registry` it's available in the Justfile! it's not a flow that the users of this repo need to run so we don't need to add it to README i think

yep! thanks! :)

PR #370, opened 2025-04-03
fix: add lookup entry to access list generation for relayMessage txs

tremarkley:
Updates supersim to use a conversion function provided by the `op-supervisor` package to encode the Access List for a relayMessage tx. Previously the way supersim encoded the Access List, did not include a lookup entry. While this would work within supersim, a lookup entry is required when relaying a message using public infra. 

Note from proto:
> To clarify: the "lookup entry" part, type 1, is a bytes32 in the access-list that defines what log-event is being accessed, so no execution is needed to determine where to look for the data to check the checksum against. It's required if you post messages to public infra that does these checks ahead of EVM execution. During direct execution, if you're able to skip the pre-EVM checks (e.g. in foundry), the checksum entry (type 3) alone is sufficient.

Since genesis files have to be updated in order to make this change, I also updated the `DeployL2PeripheryContracts` to etch in the Promise contract at the expected address.


tremarkley:
* **#370** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/370?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/370?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tarunkhasnavis (approved):

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "contracts/lib/optimism"]
 	path = contracts/lib/optimism
 	url = https://github.com/ethereum-optimism/optimism
+[submodule "contracts/lib/interop-lib"]
```
fainashalts, line 7: adding a submodule increases repository management complexity. do you think we could add this as a Foundry dependency instead? there's probably a good reason not to but wanted to raise it as an option to see what you think!

fainashalts: actually I take it back. since interop-lib probably won't be stable for a while as changes are being made submodule gives us better control.

tremarkley: oh interesting, is there a way to install foundry dependencies without using sub modules? [`forge install`'s](https://book.getfoundry.sh/reference/forge/forge-install) default behavior is to add the dependency as a submodule and I think the only other option outside of submodules is to use node_modules, but definitely open to other options if it's possible!

fainashalts: Yeah you can ignore me. :)

PR #373, opened 2025-04-07
prototype: phantom superchain erc20

hamdiallam:


hamdiallam:
* **#373** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/373?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/supersim/373?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

contracts/src/PhantomSuperchainERC20.sol
```diff
@@ -0,0 +1,113 @@
...
+import {console} from "forge-std/console.sol";
+
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {Create2} from "@openzeppelin/contracts/utils/Create2.sol";
+
+import {IL2ToL2CrossDomainMessenger} from "@contracts-bedrock-interfaces/L2/IL2ToL2CrossDomainMessenger.sol";
+import {Predeploys} from "@contracts-bedrock/libraries/Predeploys.sol";
+
+contract PhantomSuperchainERC20 is ERC20 {
```
karlfloersch, line 12: I worry that this ERC20 is too feature rich and will cause foot guns. For example things that won't work well:
* Approve
* TotalSupply

because they will all return values which seem correct but actually are broken. If you give someone an API, they will use it even if you don't intend for them to use it.

I think a way to adjust this is to actually revert if anything is called that we can't support 1:1 with an ERC20. That way, when integrating this into an existing application we can also get noisy errors if someone integrates the phantom and then their tests fail because they use `approve`

Oh another note -- the API we support is actually a bit nuanced because we don't actually fully support transferFrom or balanceOf. We ONLY support `transferFrom(addr)` or `balanceOf(addr)` if the addr==the contract which is using the token. So we should enforce that behavior as well

hamdiallam: yea 100%, inheriting ERC20 here was just to bootstrap this implementation and not concern myself with having erc20 stuff implemented.

> `transferFrom`: agreed we should not implement this. However this API could by default be the `crossDomainDepositFrom` in the context of the phantom token to remotely transfer the funds into a recipient address. Kinda fits the name nicely. Only the messenger would be authorized as `msg.sender`.

++1 on approvals and such

PR #396, opened 2025-06-09
feat: GasTank deployment

tremarkley:


contracts/script/DeployL2PeripheryContracts.s.sol
```diff
@@ -44,19 +46,9 @@ contract DeployL2PeripheryContracts is Script {
         // contract to this address in devnet, so we need to manually etch the code to the
         // expected address
         vm.etch(PredeployAddresses.PROMISE, p.code);
-        // Number of slots to check
-        uint256 slotsToCheck = 1000;
 
-        // Copy storage slots
-        for (uint256 slot = 0; slot < slotsToCheck; slot++) {
-            bytes32 value = vm.load(p, bytes32(slot));
+        copyStorageSlots(p, PredeployAddresses.PROMISE, 1000);
```
hamdiallam, line 50: I was recently looking at this. Do we need to copy any storage slots? it should all be empty at genesis anyways. I think we can just delete this

hamdiallam (approved):

contracts/script/DeployL2PeripheryContracts.s.sol
tremarkley: there is a storage slot which holds pointer to the messenger contract: https://github.com/ethereum-optimism/interop-lib/blob/main/src/Promise.sol. it might be overkill for this to check 1000 slots though lol

PR #398, opened 2025-06-11
feat: update `GasTank` to latest design

tremarkley:
Updates the `GasTank` deployed at genesis to the latest iteration of the `GasTank` contract: https://github.com/defi-wonderland/optimism/pull/414/files


artifact/artifact.go
```diff
@@ -15,7 +15,7 @@ type Bytecode struct {
 }
 
 type artifactMarshaling struct {
-	Bytecode Bytecode `json:"bytecode"`
+	Bytecode Bytecode `json:"deployedBytecode"`
```
its-everdred, line 18: ❓Does this functionally change the type since they're both just hex strings?

tremarkley: sorry I should have wrote a comment about this in the PR description. this needed to be updated because when setting the bytecode for the contract, it needs to use the `deployedBytecode` field instead of the `bytecode` field, however, both fields have the same type so the type is the same

Issue #361, opened 2025-03-21
Configurable Dependency Set


tremarkley:
Right now all L2 chains in supersim are in the same dependency set. This issue is for allowing the dependency set to be configurable via a cli flag passed into supersim on startup.

fainashalts:
Great idea! 👍 Especially useful if someone wants to test out an app that works in and outside of the dependency set.

tremarkley:
I think you could potentially just have a flag that is `--dependency-set <chains>` and any chains set in the flag would be in a dependency set together. This is just me thinking off the top of my head, so there may be a more ideal format. But basically you want to have a way to specify what the dependency set looks like for a set of supersim chains. Here is the spec for dependency set: https://specs.optimism.io/interop/dependency-set.html and you can grep the monorepo to see how the dependency set is managed, this might be a good place to start: https://github.com/ethereum-optimism/optimism/blob/develop/op-program/client/tasks/derive.go#L40

PR #407, opened 2025-06-20
Configurable dependency set

its-everdred:
## Problem
[ticket](https://github.com/ethereum-optimism/supersim/issues/361)
At the moment, there is no way for `supersim` to configure a custom dependency set.

## Solution
Add a flag:  `--dependency.set` that takes either `[1,2,3]` or `1,2,3`
Chain IDs passed are appended to the existing set. For example:
`supersim --dependency.set 8453`
will append and log the sets:
```
...
Chain Configuration
-----------------------
L1: Name: Local  ChainID: 900  ...

L2: Predeploy Contracts Spec

  * Name: OPChainA  ChainID: 901 ...
    Dependency Set: [902, 8453]
    L1 Contracts:
     - OptimismPortal:         ...
     ...

  * Name: OPChainB  ChainID: 902 ...
    Dependency Set: [901, 8453]
    L1 Contracts:
     - OptimismPortal:         ...
     ...
```


config/chain.go
```diff
@@ -166,16 +166,23 @@ func GetNetworkConfig(cliConfig *CLIConfig) NetworkConfig {
 			InteropL2ToL2CDMOverrideArtifactPath: cliConfig.InteropL2ToL2CDMOverrideArtifactPath,
 		}
 
-		// populate dep set
+		// populate dep set with local chains
 		for j := uint64(0); j < cliConfig.L2Count; j++ {
```
its-everdred, line 170: ⚠️ `uint64(0)` is technically not to [spec](https://specs.optimism.io/interop/dependency-set.html#chain-id), chainIDs can theoretically be as large as `uint256`. This is baked into a lot of places, I can either use this PR to change everywhere or create a separate issue.

opsimulator/opsimulator.go
```diff
@@ -510,7 +515,22 @@ func (opSim *OpSimulator) checkInteropInvariants(ctx context.Context, logs []typ
...
+	executingChainDependencySet := opSim.Config().L2Config.DependencySet
+
+	// Check if source chain is in executing chain's dependency set
+	for _, depChainID := range executingChainDependencySet {
+		if depChainID == sourceChainID {
+			return nil // Source chain is in dependency set
+		}
+	}
+
+	return fmt.Errorf("executing message in block (chain %d) may not execute message from chain %d: not in dependency set", executingChainID, sourceChainID)
```
its-everdred, line 530: 🤔 Does throwing this error create a false sense security, or is it beneficial to dev testing?

tremarkley: I think this is good. it is helpful because in prod the tx will be dropped, so end behavior is still the same, this just adds additional logging.

orchestrator/orchestrator.go
```diff
@@ -270,7 +270,16 @@ func (o *Orchestrator) ConfigAsString() string {
...
+				fmt.Fprintf(&b, "  * Name: %s  ChainID: %d  RPC: %s  LogPath: %s\n", cfg.Name, cfg.ChainID, opSim.Endpoint(), opSim.LogPath())
+
+		if len(cfg.L2Config.DependencySet) > 0 {
+			depSetStrs := make([]string, len(cfg.L2Config.DependencySet))
+			for i, chainID := range cfg.L2Config.DependencySet {
+				depSetStrs[i] = fmt.Sprintf("%d", chainID)
+			}
+			fmt.Fprintf(&b, "    Dependency Set: [%s]\n", strings.Join(depSetStrs, ", "))
+		}
+
```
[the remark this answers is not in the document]
tremarkley: nice

opsimulator/opsimulator_test.go
```diff
@@ -0,0 +1,243 @@
...
+		Raw: types.Log{
+			Address:     predeploys.CrossL2InboxAddr,
+			Topics:      []common.Hash{crypto.Keccak256Hash([]byte("ExecutingMessage"))},
+			Data:        []byte("mock data"),
+			BlockNumber: 100,
+		},
+	}
+}
+
+func TestCheckInteropInvariants_DependencySetValidation(t *testing.T) {
```
tremarkley, line 68: would be good to add a test case that validates that a message with same source and destination, passes validation

config/chain_test.go
```diff
@@ -0,0 +1,110 @@
...
+		depSet := l2Config.L2Config.DependencySet
+		require.NotContains(t, depSet, l2Config.ChainID,
+			"Chain %d should never include itself in dependency set", l2Config.ChainID)
+	}
+}
+
+func TestGetNetworkConfig_EmptyDependencySet(t *testing.T) {
+	// Test with empty user dependency set
+	cliConfig := &CLIConfig{
+		L2Count:       1,
```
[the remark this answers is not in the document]
its-everdred: done!

config/cli_test.go
```diff
@@ -0,0 +1,399 @@
...
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tt.expectedError)
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestCLIConfig_Check_WithDependencySet(t *testing.T) {
```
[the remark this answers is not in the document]
its-everdred: DRYed this file up a lot

opsimulator/opsimulator_test.go
```diff
@@ -0,0 +1,197 @@
...
+			name:                   "Valid: Multiple chains in dependency set",
+			executingChainID:       903,
+			executingDependencySet: []uint64{901, 902},
+			sourceChainID:          901,
+			expectError:            false,
+		},
+		{
+			name:                   "Valid: Same-chain message (source == destination)",
+			executingChainID:       901,
+			executingDependencySet: []uint64{902}, // 901 is NOT in its own dependency set
```
tremarkley, line 83: nit: this comment confused me a little bit since even though 901 is not in its own dependency set, it is still a valid message. it could be good to remove it, unless you think it is helpful.

opsimulator/opsimulator.go
```diff
@@ -510,7 +515,28 @@ func (opSim *OpSimulator) checkInteropInvariants(ctx context.Context, logs []typ
 	return nil
 }
 
-// Overridden such that the port field can appropiately be set
+// ValidateDependencySet checks if the source chain is allowed to send messages to the executing chain
+func (opSim *OpSimulator) ValidateDependencySet(sourceChainID uint64) error {
```
tremarkley, line 519: similar comment here, it would be preferable to keep this function internal to the opsimulator since it is not needed by other packages, unless there is a good reason for it being public. Sorry for not catching this on first review.

supersim_test.go
```diff
@@ -1301,3 +1302,114 @@ func TestEthSubscribeNewHeads(t *testing.T) {
...
+			destChain := testSuite.Supersim.Orchestrator.L2Chains()[0]
+			for _, chain := range testSuite.Supersim.Orchestrator.L2Chains() {
+				if chain.Config().ChainID == tc.destChainID {
+					destChain = chain
+					break
+				}
+			}
+
+			// Test the dependency set validation logic
+			err := destChain.(*opsimulator.OpSimulator).ValidateDependencySet(tc.sourceChainID)
```
tremarkley, line 1405: instead of calling `ValidateDependencySet` directly, it would be better to test this fully end to end by attempting to send a message to a chain outside the dependency set. You can use these tests as references for how to send a crosschain message: [test for manually relaying a crossdomain message](https://github.com/ethereum-optimism/supersim/blob/88dcdcee5efa72d5def5b35a39f09b70a5e860d6/supersim_test.go#L603), [test for crossdomain eth bridging with autorelay mode on](https://github.com/ethereum-optimism/supersim/blob/88dcdcee5efa72d5def5b35a39f09b70a5e860d6/supersim_test.go#L1005),

config/cli.go
```diff
@@ -283,3 +371,47 @@ func validateHost(host string) error {
 
 	return nil
 }
+
+func ParseDependencySet(dependencySet string) ([]uint64, error) {
```
tremarkley, line 375: does this have to be public or can it be kept internal to the cli? also if you want to use this for a unit test, I believe that you can still call it as long as the unit test has its package set to `package config`. Sorry I didnt catch this the first time around

opsimulator/opsimulator_test.go
its-everdred: makes sense, removed.

opsimulator/opsimulator.go
its-everdred: done!

tremarkley (approved):
well done!

